PR #18, opened 2023-04-04
Use SPI to store quay credentials of Component image repository

mmorhun:
This PR makes Image Controller store generated quay credentials into SPI and then use SPI API to provide the credentials to the pipeline. Additionally, SPI is used to bind the credentials to pipeline Service Account.

Beside image repository provision flow, regeneration of credentials (with following update in SPI) is implemented. To rotate generated quay credentials, set `image.redhat.com/generate` annotation to `regenerate-token`.

Since SPI could change internal storage that causes removal of all stored in SPI tokens, we need to make sure, that the Component image repository credentials are in a valid state. To mitigate that - not often, but possible - case, there is a script that would trigger token objects recreation:
```bash
#!/bin/bash

echo
echo "Starting Component image repository credentials recreation"
echo

tokenList=$(oc get SPIAccessTokens --all-namespaces -o json)
tokens=$(echo "$tokenList" | jq -c '.items[]')
for token in $tokens; do
    # Filter SPIAccessTokens not related to any Component
    ownerKind=$(echo "$token" | jq -r '.metadata.ownerReferences[0].kind')
    if [ "$ownerKind" != "Component" ]; then        
        continue
    fi
    
    # Filter SPIAccessTokens in good condition
    tokenPhase=$(echo "$token" | jq -r '.status.phase')
    if [ "$tokenPhase" == "Ready" ]; then        
        continue
    fi
    
    componentName=$(echo "$token" | jq -r '.metadata.ownerReferences[0].name')
    namespace=$(echo "$token" | jq -r '.metadata.namespace')
    tokenName=$(echo "$token" | jq -r '.metadata.name')

    # Clean old token and binding
    oc delete SPIAccessTokenBinding "$tokenName" -n "$namespace"
    oc delete SPIAccessToken "$tokenName" -n "$namespace"
    # Do image repository reprovision
    oc annotate Component "$componentName" -n "$namespace" image.redhat.com/generate=true --overwrite    
    echo "Requested token recreation for $componentName Component in $namespace namespace"
    echo
done

echo "Component image repository credentials recreation is done"
echo
```
Run the script above right after SPI viped all tokens from its internal storage.


mmorhun:
@skabashnyuk, normally, it's not possible for the described use-case because we control the token flow (creation/deletion) and robot account token doesn't have expiration date. Also users don't have access to the organization to invalidate the token.
If a token was leaked, then the token is still valid, but needs to be refreshed, which is already implemented in this PR.
Do you have any other concerns?

controllers/component_image_controller.go
```diff
@@ -50,20 +52,47 @@ type RepositoryInfo struct {
 	Secret string `json:"secret"`
 }
 
+type ImageControllerAction int
+
+const (
+	NoAction ImageControllerAction = iota
+	ProvisionImageRepositoryAction
+	RegenrateImageRepositoryTokenAction
```
Michkov, line 60: typo, Regenrate -> Regenerate

Michkov (commented):
Update README.md

mmorhun:
Refreshing quay token is broken now due to [this issue](https://issues.redhat.com/browse/SVPI-441).

mmorhun:
@skabashnyuk updated the PR description with the script that solves problems on SPI backend change.

README.md
```diff
@@ -41,6 +41,10 @@ spec:
 
 If `Component`'s auto-generated image repository should be deleted after component deletion, add `image.redhat.com/delete-image-repo` annotation to the `Component`.
 
+### Update image repository token of the Component
+
+To refresh image repository access token, set annotation `image.redhat.com/generate: 'regenerate-token'` on the Component.
```
Michkov, line 46: also rest of readme should be checked, eg. now there will not be secret in `image.redhat.com/image` annotation

tkdchen (approved):

Michkov (changes requested):
The generated secret is:
```
	"auths": {
		"quay.io": {
			"auth": "XXXX"
		}
	}
```
we need to be scoped to repo:
```
	"auths": {
		"quay.io/mkovarik-test/spi-test/test-application/devfile-sample-python-basic": {
			"auth": "XXXX"
		}
	}
```

mmorhun:
Resolved conflicts and rebased

controllers/component_image_controller.go
```diff
@@ -142,150 +164,354 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	generatedRepository := RepositoryInfo{
-		Image:  imageURL,
-		Secret: robotAccountSecret.Name,
+	// Add finalizer to the component in order to clean up robot account and image repository if requested
+	if !controllerutil.ContainsFinalizer(component, ImageRepositoryFinalizer) && component.ObjectMeta.DeletionTimestamp.IsZero() {
+		if err := r.Client.Get(ctx, componentKey, component); err != nil {
+			log.Error(err, "failed to get component")
+			return false, err
+		}
+		if !controllerutil.ContainsFinalizer(component, ImageRepositoryFinalizer) && component.ObjectMeta.DeletionTimestamp.IsZero() {
```
tkdchen, line 287: Why check this twice?

mmorhun: Because component could get updates. We reread its configuration right above.

mmorhun:
Applied newly implemented SPI feature to provide docker config JSON in Kubernetes format.
Now, multi-component application build works:
image

controllers/component_image_controller.go
```diff
@@ -142,150 +164,367 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			r.QuayClient.DeleteRepository(r.QuayOrganization, imageRepoName)
+			return false, fmt.Errorf("image repository provision canceled due to component deletion")
+		}
 		if !controllerutil.ContainsFinalizer(component, ImageRepositoryFinalizer) {
 			controllerutil.AddFinalizer(component, ImageRepositoryFinalizer)
+			if err := r.Client.Update(ctx, component); err != nil {
+				log.Error(err, "failed to update component")
+				return false, err
+			}
+			log.Info("Image regipository finaliziler added to the Component")
```
Michkov, line 304: repository finalizer

mmorhun:
Blocked by another SPI issue

mmorhun:
Implemented in [another PR](https://github.com/redhat-appstudio/image-controller/pull/40).

PR #21, opened 2023-04-17
feat: Handle robot account longname in quay

tnevrlka:
`CreateRobotAccount` and `DeleteRobotAccount` may fail when accidentally using longform robot account name because [quay.io api](https://docs.quay.io/api/swagger) requires shortname. 
Added `handleRobotName` which will convert longname to shortname and leave shortname be.
e.g. `org+robot` will be changed to `robot`, `robot` will stay `robot`

### DIFF
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -385,6 +385,7 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 func handleRobotName(robotName string) (string, error) {
 	// Regexp from quay api `^([a-z0-9]+(?:[._-][a-z0-9]+)*)$` with one plus sign in the middle allowed (representing longname)
 	r, err := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
+	robotName = strings.TrimSpace(robotName)
 	if err != nil {
 		return "", fmt.Errorf("failed to compile regex, error: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/quay/quay.go
```diff
@@ -370,3 +378,15 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 	}
 	return response.Robots, nil
 }
+
+// If robotName is in longform, return shortname
+func handleRobotName(robotName string) (string, error) {
+	if strings.Contains(robotName, "+") {
+		parts := strings.Split(robotName, "+")
+		if len(parts) != 2 {
+			return robotName, fmt.Errorf("robotName could not be split into two parts, expected len 2, got len %d", len(parts))
```
psturc, line 387: I think in case of an error, empty string should be returned

pkg/quay/quay.go
```diff
@@ -370,3 +378,15 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 	}
 	return response.Robots, nil
 }
+
+// If robotName is in longform, return shortname
+func handleRobotName(robotName string) (string, error) {
```
psturc, line 383: thinking if we could do (as an extra input validation) a check if:
* the name is empty 
* it contains invalid robot name like "org+"

and those cases could be added to unit tests

psturc (commented):
minor suggestions, otherwise lgtm

pkg/quay/quay.go
tnevrlka: Maybe use the regexp provided by the quay api with the exception of the longname form?

tisutisu (commented):
LGTM

psturc (approved):

PR #22, opened 2023-04-19
feat: Add functions for getting and deleting quay tags

tnevrlka:
Added functions for getting and deleting quay tags for use in [e2e-tests cleanup](https://github.com/redhat-appstudio/e2e-tests/pull/467)
Covered by unit tests, tested in my own quay org

### DIFF
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -35,6 +35,8 @@ type QuayService interface {
 	AddPermissionsToRobotAccount(organization, imageRepository, robotAccountName string) error
 	GetAllRepositories(organization string) ([]Repository, error)
 	GetAllRobotAccounts(organization string) ([]RobotAccount, error)
+	GetTagsFromPage(organization, repository string, page int) ([]Tag, bool, error)
+	DeleteTag(organization, repository, tag string) (bool, error)
 }
 
 var _ QuayService = (*QuayClient)(nil)
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -464,5 +464,18 @@ func (c *QuayClient) DeleteTag(organization, repository, tag string) (bool, erro
 	if res.StatusCode == 404 {
 		return false, nil
 	}
-	return false, fmt.Errorf("error deleting tag")
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return false, fmt.Errorf("failed to read response body, error: %s", err)
+	}
+	data := &QuayError{}
+	err = json.Unmarshal(body, data)
+	if err != nil {
+		return false, fmt.Errorf("failed to unmarshal body, error: %s", err)
+	}
+	if data.Error != "" {
+		return false, errors.New(data.Error)
+	}
+	return false, errors.New(data.ErrorMessage)
 }
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -432,7 +432,7 @@ func (c *QuayClient) GetTagsFromPage(organization, repository string, page int) 
 	var response struct {
 		Tags          []Tag `json:"tags"`
 		Page          int   `json:"page"`
-		HasAdditional bool  `json:"has_additonal"`
+		HasAdditional bool  `json:"has_additional"`
 	}
 	err = json.Unmarshal(body, &response)
 	if err != nil {
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -401,3 +401,66 @@ func handleRobotName(robotName string) (string, error) {
 	}
 	return robotName, nil
 }
+
+func (c *QuayClient) GetTagsFromPage(organization, repository string, page int) ([]Tag, bool, error) {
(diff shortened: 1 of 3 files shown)


pkg/quay/quay.go
```diff
@@ -401,3 +401,66 @@ func handleRobotName(robotName string) (string, error) {
...
+		return false, fmt.Errorf("failed to Do request, error: %s", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 204 {
+		return true, nil
+	}
+	if res.StatusCode == 404 {
+		return false, nil
+	}
```
tkdchen, line 466: The final result of deleting a tag is the tag does not exist any more. When Quay.io responds 404, isn't it a kind of expectation? Or, 404 has other special meaning in this context?

pkg/quay/quay_test.go
```diff
@@ -262,3 +263,132 @@ func TestQuayClient_handleRobotName(t *testing.T) {
...
+	defer gock.Off()
+
+	org := "test_org"
+	repo := "test_repo"
+
+	testCases := []struct {
+		name          string
+		pages         int
+		tagsPerPage   int
+		totalTags     int
```
tisutisu, line 277: Do we require `totalTags` to be here?

tnevrlka: You're right, it's no longer necessary. Thank you

pkg/quay/quay.go
tnevrlka: I see your point, I think it might be a matter of personal preference. 
My reasons for leaving it in are:
- I believe it is more flexible, off the top of my head, I can imagine wanting to log if the tag was actually deleted by the function or if it was not found.
- Its behavior is consistent with DeleteRepository and DeleteRobotAccounts functions, so those would have to be changed as well.

pkg/quay/quay.go
```diff
@@ -401,3 +403,66 @@ func handleRobotName(robotName string) (string, error) {
...
+	return false, fmt.Errorf("error deleting tag")
```
psturc, line 467: should this error be more specific? i.e. similarly [how it's done in DeleteRepository method](https://github.com/redhat-appstudio/image-controller/pull/22/files#diff-d4a36e5f4e14c5636dfed4031b618cc2a45afc79934eabb0336a1adde773fd1bL126-R137)

pkg/quay/quay.go
```diff
@@ -464,5 +464,18 @@ func (c *QuayClient) DeleteTag(organization, repository, tag string) (bool, erro
...
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return false, fmt.Errorf("failed to read response body, error: %s", err)
+	}
+	data := &QuayError{}
+	err = json.Unmarshal(body, data)
+	if err != nil {
+		return false, fmt.Errorf("failed to unmarshal body, error: %s", err)
+	}
+	if data.Error != "" {
```
psturc, line 477: it is super strange that quay sometimes returns content in `error_message` and sometimes in `error`. good spot thought

would it be worth to add this condition to other methods as well? for example I noticed that in case of an error in `DeleteRepository` method the error is present in the `error` field as well, not in `error_message`, as we're returning [here](https://github.com/redhat-appstudio/image-controller/pull/22/commits/c31e8607e77f16033914a353493028014eadf9b0#diff-d4a36e5f4e14c5636dfed4031b618cc2a45afc79934eabb0336a1adde773fd1bL137)
Screenshot 2023-04-21 at 13 12 03

psturc (commented):
just one question/suggestion, otherwise lgtm 👍

pkg/quay/quay.go
tnevrlka: Unless there is a nicer way to handle both possibilities, I can do it, sure

psturc: probably worth to check whether there's even some API endpoint that we're calling in these methods where we can expect `error_message` field populated in case of an error. maybe we don't need `error_message` field at all

pkg/quay/quay.go
```diff
@@ -401,3 +403,79 @@ func handleRobotName(robotName string) (string, error) {
...
+	defer res.Body.Close()
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to read response body, error: %s", err)
+	}
+
+	var response struct {
+		Tags          []Tag
+		Page          int
+		HasAdditional bool
```
psturc, line 435: I think these fields should have json struct tags, at least definitely `HasAdditional`
```go
HasAdditional bool `json:"has_additional"`
```
otherwise the response is not unmarshalled correctly

psturc (changes requested):

pkg/quay/quay.go
psturc: if it's simple enough to fix it (which in this case I think it is), I'd do it in this PR, just a separate commit

psturc (approved):
/lgtm

PR #24, opened 2023-05-08
feat: Added IsRepositoryExists routine 

tisutisu:
Added routine `IsRepositoryExists()` in the quay package which we require in our e2e-tests to validate if the repository exists created by image-controller.  Change also contains two debug tests one for checking repository and another is for checking RobotAccount.

### DIFF
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -96,6 +96,44 @@ func (c *QuayClient) CreateRepository(r RepositoryRequest) (*Repository, error) 
 	return data, nil
 }
 
+// IsRepositoryExists checks if the specified image repository exists in quay.
+func (c *QuayClient) IsRepositoryExists(organization, imageRepository string) (bool, error) {
+	url := fmt.Sprintf("%s/repository/%s/%s", c.url, organization, imageRepository)
+
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return false, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Add("Authorization", fmt.Sprintf("%s %s", "Bearer", c.AuthToken))
+	req.Header.Add("Content-Type", "application/json")
+
+	res, err := c.httpClient.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("failed to Do request: %w", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 404 {
+		return false, fmt.Errorf("Not Found")
+	} else if res.StatusCode == 200 {
+		return true, nil
+	}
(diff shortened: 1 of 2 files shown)


pkg/quay/quay.go
```diff
@@ -96,6 +96,44 @@ func (c *QuayClient) CreateRepository(r RepositoryRequest) (*Repository, error)
 	return data, nil
 }
 
+// IsRepositoryExists checks if the specified image repository exists in quay.
+func (c *QuayClient) IsRepositoryExists(organization, imageRepository string) (bool, error) {
```
psturc, line 100: why not just `RepositoryExists`?

tnevrlka, line 100: Consider renaming the function to `DoesRepositoryExist` or `RepositoryExists`.
I would personally prefer `DoesRepositoryExist`.

pkg/quay/quay.go
```diff
@@ -96,6 +96,44 @@ func (c *QuayClient) CreateRepository(r RepositoryRequest) (*Repository, error)
...
+	req.Header.Add("Content-Type", "application/json")
+
+	res, err := c.httpClient.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("failed to Do request: %w", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == 404 {
+		return false, fmt.Errorf("Not Found")
```
psturc, line 117: could we add more info to the error message here? (org, image repository)

tisutisu: Kept it short intentionally so that when we use this routine in our e2e-tests, we will compare this returned error message with `Not Found` string to confirm when the repo does not exists, similar to [here](https://github.com/redhat-appstudio/image-controller/pull/24/files#diff-f49653252dd0f96d650015e2ebd690a20d9e7872dc1ebefc9c671da72d6a9f19R63). If we add additional info, that will make comparison complex. WDYT?

pkg/quay/quay.go
tisutisu: Changed it to `DoesRepositoryExist` now. To make it consistent with other routine name that are starting with a `verb` like `Create` `Delete` `Get`.

pkg/quay/quay.go
psturc: my point was that when the error is hit, it's good when it's as informative as possible. 

> If we add additional info, that will make comparison complex. WDYT?

instead of checking the whole error, could we check its prefix instead, like `strings.HasPrefix()`, wdyt

psturc:
@tnevrlka @tisutisu  my point: `if RepositoryExists()` vs `if DoesRepositoryExist()` 😛

PR #25, opened 2023-05-18
added github action workflows for linting, gosec and codecov

tisutisu:
This PR will address story: https://issues.redhat.com/browse/STONEBLD-1192

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,0 +1,60 @@
+name: Validate PR
+on:
+  pull_request:
+    branches: [ main ]
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: './go.mod'
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          args: --timeout=5m
+
+  security_scan:
+    name: Security scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: './go.mod'
+      - name: Run tests and collect coverage
+        run: make test
+      - name: Upload codecov report
+        uses: codecov/codecov-action@v3.1.1
+        with:
+          fail_ci_if_error: true
```
tnevrlka, line 59: Would it make sense to split this into two jobs like [here](https://github.com/redhat-appstudio/build-service/blob/6588316a6842af5276ab13650702f37d8b7c2568/.github/workflows/pr.yaml)?

Might not be a problem anymore, but I previously encountered problems with uploads to codecov.
If the codecov upload fails, it would require rerunning the entire job, which be time-consuming if `make test` takes several minutes. Rerunning codecov upload takes max 10 seconds.
The split might also simplify the debugging process since we would know if the error occured in the tests or codecov

tisutisu: `I previously encountered problems with uploads to codecov` could you please specify what was the exact problem that you encountered? I don't think splinting into two separate jobs one for running the test and another for codecov upload is needed. In case go test step fails, we can debug that from the github actions logs itself. I think the upload artifacts and download artifacts in the build-service github actions overkill.

tisutisu: WDYT?

tnevrlka: > could you please specify what was the exact problem that you encountered?

I wasn't able to find it. Anyway I took a look at some recent logs and the upload doesn't seem to fail anymore, so not an issue.

I believe you're right, debugging from logs is fine, so probably no reason for the split.

tnevrlka (approved):
LGTM

psturc (approved):

PR #26, opened 2023-06-05
Retry HTTP requests to Quay.io

tkdchen:
This patch introduce a retry mechanism at the HTTP level for quay package to interace with Quay.io API. It affects every API defined in quay package.

go-retryablehttp package is used and does the actual retries underlying. Retry only happens when connection error occurs and Quay.io responds 5xx response code.


tkdchen:
Regarding the Lint job, requires #27

Michkov (approved):

mmorhun (approved):

PR #30, opened 2023-06-09
RHTAPBUGS-371: Wait for cache update

Michkov:
Wait till the cache with Component CR is updated with new annotation to avoid re-running reconcile loop on non-updated Component CR.

### DIFF
--- a/controllers/component_image_controller.go
+++ b/controllers/component_image_controller.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


controllers/component_image_controller.go
```diff
@@ -201,13 +201,42 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					break
+				}
+				// Outdated version of the component, wait more.
+			} else {
+				if errors.IsNotFound(err) {
+					// The component was deleted
+					isComponentInCacheUpToDate = true
+					break
+				}
+				log.Error(err, "failed to get the component")
```
tkdchen, line 233: Would be good to give a context of this message

tkdchen (approved):

controllers/component_image_controller.go
Michkov: updated

PR #32, opened 2023-06-15
Prune sbom and attestation images from the registry

mkosiarc:
If application image is removed from quay, we need to make sure that the associated sbom and attestation images are removed as well.

This is done with cronjob that will run python script that checks registry repositories via API.

[STONEBLD-1251](https://issues.redhat.com//browse/STONEBLD-1251)

### DIFF
--- a/config/registry_image_pruner/cronjob.yaml
+++ b/config/registry_image_pruner/cronjob.yaml
@@ -1,0 +1,43 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: image-pruner-cronjob
+spec:
+  schedule: "0 0 * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: image-pruner
+            image: registry.redhat.io/rhel8/python-39:1-120.1684740828
+            env:
+              - name: QUAY_TOKEN
+                valueFrom:
+                  secretKeyRef:
+                    name: quaytoken
+                    key: quaytoken
+              - name: NAMESPACE
+                valueFrom:
+                  secretKeyRef:
+                    name: quaytoken
+                    key: organization
+            imagePullPolicy: IfNotPresent
+            command:
+              - /bin/bash
+              - '-c'
+              - >-
+                pip install -r /image-pruner/requirements.txt && python
+                /image-pruner/prune_images.py --namespace $(NAMESPACE)
+            volumeMounts:
+              - name: image-pruner-volume
+                mountPath: /image-pruner
+          restartPolicy: OnFailure
+          volumes:
+            - name: image-pruner-volume
+              configMap:
+                name: image-pruner-configmap
+            - name: quaytoken
+              secret:
+                secretName: quaytoken
--- a/config/registry_image_pruner/kustomization.yaml
+++ b/config/registry_image_pruner/kustomization.yaml
@@ -1,0 +1,8 @@
+resources:
+- cronjob.yaml
+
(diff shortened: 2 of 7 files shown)


config/registry_image_pruner/cronjob.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: image-pruner-cronjob
+spec:
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: image-pruner
+            image: registry.redhat.io/rhel8/python-39
```
Michkov, line 13: bind to tag and sha

config/registry_image_pruner/cronjob.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: batch/v1
+kind: CronJob
```
Michkov, line 2: maybe we want to disable concurency

config/registry_image_pruner/cronjob.yaml
```diff
@@ -0,0 +1,42 @@
...
+                  secretKeyRef:
+                    name: quaytoken
+                    key: organization
+            imagePullPolicy: IfNotPresent
+            command:
+              - /bin/bash
+              - '-c'
+              - >-
+                pip install -r /image-pruner/requirements.txt && python
+                /image-pruner/prune_images.py --namespace $(NAMESPACE) --dry-run
```
Michkov, line 31: dry run not needed

config/registry_image_pruner/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+namespace: image-controller-system
+
```
Michkov, line 2: include this file in config/base

mkosiarc: What do you mean? I don't see config/base

config/registry_image_pruner/cronjob.yaml
mkosiarc: updated

config/registry_image_pruner/cronjob.yaml
mkosiarc: was still leaving that for testing purposes, but updated.

config/registry_image_pruner/cronjob.yaml
mkosiarc: binded to the newest version tag available

mmorhun (approved):
LGTM
Please add missing new lines at the end of files

config/registry_image_pruner/kustomization.yaml
mmorhun: I think that `image-controller-system` should be "kustomizable".
In real clusters, usually, the namespace is `image-controller`.

mkosiarc: I am still lost :D

Michkov: Ok I meant https://github.com/redhat-appstudio/image-controller/blob/main/config/default/kustomization.yaml

Michkov: add `../registry_image_pruner` into bases and test by running `kustomize build config/default` that should generate the CRs also with the cronjob.

mkosiarc: I did that and also removed namespace and nameprefix from config/registry_image_pruner/kustomization.yaml as they are mentioned in the default and other kustomization files don't have them.

Michkov (approved):

PR #35, opened 2023-07-04
Read configuration at runtime

filariow:
The controller is loading the Quay's token data at startup, so it doesn't load secret's updates.
When a Quay token is rotated, the new values are projected into the /workspace folder.
However, the controller will continue using the one loaded at startup, and start gracefully handling 403 replies from Quay.
This implies that to effectively rotate the Quay's token you also need to restart the controller.

With this patch the token data is retrieved when needed instead of at startup, meaning that pod restart should not be required when rotating Quay's token.

### DIFF
--- a/controllers/component_image_controller.go
+++ b/controllers/component_image_controller.go
@@ -76,25 +76,6 @@ type ComponentReconciler struct {
 	QuayOrganization string
 }
 
-type LazyQuayClient struct {
-	buildQuayClient func() quay.QuayService
-	quayClient      quay.QuayService
-}
-
-func NewLazyQuayClient(buildQuayClient func() quay.QuayService) *LazyQuayClient {
-	return &LazyQuayClient{
-		buildQuayClient: buildQuayClient,
-		quayClient:      nil,
-	}
-}
-
-func (l *LazyQuayClient) QuayClient() quay.QuayService {
-	if l.quayClient == nil {
-		l.quayClient = l.buildQuayClient()
-	}
-	return l.quayClient
-}
-
 // SetupWithManager sets up the controller with the Manager.
 func (r *ComponentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
@@ -224,7 +205,6 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	// Do something only if no error has been detected before
 	if repositoryInfo.Message == "" {
-		lazyQuayClient := NewLazyQuayClient(r.BuildQuayClient)
 		if imageRepositoryExists {
 			// Check if need to change image repository  visibility
 			if repositoryInfo.Visibility != requestRepositoryOpts.Visibility {
@@ -232,7 +212,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				imageUrlParts := strings.SplitN(repositoryInfo.Image, "/", 3)
 				if len(imageUrlParts) > 2 {
 					repositoryName := imageUrlParts[2]
-					quayClient := lazyQuayClient.QuayClient()
+					quayClient := r.BuildQuayClient()
 					if err := quayClient.ChangeRepositoryVisibility(r.QuayOrganization, repositoryName, requestRepositoryOpts.Visibility); err == nil {
 						repositoryInfo.Visibility = requestRepositoryOpts.Visibility
 					} else {
@@ -250,7 +230,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			}
 		} else {
 			// Image repository doesn't exist, create it.
-			quayClient := lazyQuayClient.QuayClient()
+			quayClient := r.BuildQuayClient()
 			repo, pushRobotAccount, pullRobotAccount, err := r.generateImageRepository(ctx, quayClient, component, requestRepositoryOpts)
 			if err != nil {
 				if err.Error() == "payment required" {
--- a/controllers/component_image_controller.go
+++ b/controllers/component_image_controller.go
@@ -72,8 +72,8 @@ type ComponentReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 
-	QuayClient       quay.QuayService
-	QuayOrganization func() string
+	BuildQuayClient      func() quay.QuayService
+	ReadQuayOrganization func() string
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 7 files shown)


mmorhun:
I'm not sure that initiation of file IO on each request is a good way. I think we better go with event based approach.

stuartwdouglas:
> I'm not sure that initiation of file IO on each request is a good way. I think we better go with event based approach.

Are you really going to be creating and deleting repositories that fast that this would have any measurable effect? I very much doubt that the performance impact would be measurable, but the complexity of an event based system would be. Any additional load on the API server by watching the secret would likely to be more impactful that this in terms of overall system performance.

mmorhun:
> Are you really going to be creating and deleting repositories that fast that this would have any measurable effect?

No, it's not that fast, provision of a new image repository is relatively rare request. However, to do a new image repository provision, we need to do at least 5 requests in a row to quay, and for each of them the file will be reread. @stuartwdouglas do you think it's ok?

stuartwdouglas:
Once per reconcile loop would be fine I think, once per HTTP request is probably not ideal but in practice I doubt it would ever be an actual issue.

main.go
```diff
@@ -100,22 +106,22 @@ func main() {
...
-	}
-	quayClient := quay.NewQuayClient(&http.Client{Transport: &http.Transport{}}, strings.TrimSpace(string(tokenContent)), "https://quay.io/api/v1")
+
+	quayClient := quay.NewQuayClient(&http.Client{Transport: &http.Transport{}}, getReadConfigFunc(tokenPath), "https://quay.io/api/v1")
 	if err = (&controllers.ComponentReconciler{
 		Client:           mgr.GetClient(),
 		Scheme:           mgr.GetScheme(),
 		QuayClient:       quayClient,
-		QuayOrganization: strings.TrimSpace(string(orgContent)),
+		QuayOrganization: getReadConfigFunc(orgPath),
```
mmorhun, line 124: I think the organization should not be changed. We have to have a constant here.

mmorhun: I mean, the code is confusing.

mmorhun (commented):
@filariow in general, I have two remarks:
1. Quay organization is a constant and not supposed to be changed, please keep it as a constant.
2. Let's refresh the token using explicit function on each reconcile that interacts with quay API.

filariow:
>     1. Quay organization is a constant and not supposed to be changed, please keep it as a constant.
I can see your point here.
However, it seems to me it is not possible to force a constant organization.
If the administrator updates the organization key in the secret, they could always delete/recreate the pod to use the new organization value.

Also, with this approach you are going to have a mismatch between your configuration (what you have in the secret) and the data the controller is using.
It was a bit confusing to me, when I updated the secret after I found a typo in the organization name.

>     2. Let's refresh the token using explicit function on each reconcile that interacts with quay API.
I can work on this.

controllers/component_image_controller.go
```diff
@@ -72,8 +72,8 @@ type ComponentReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 
-	QuayClient       quay.QuayService
-	QuayOrganization string
+	BuildQuayClient      func() quay.QuayService
+	ReadQuayOrganization func() string
```
mmorhun, line 76: You still read the file two times in a row

controllers/component_image_controller.go
```diff
@@ -107,11 +107,14 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("error reading component: %w", err)
 	}
 
+	quayClient := r.BuildQuayClient()
+	org := r.ReadQuayOrganization()
```
mmorhun, line 111: This reads the data on each reconcile even that doesn't interact with quay.
Since we need to have new token for the deletion, I suggest to read it in two places: at the beginning of the deletion logic and below in regular reconcilation logic where we are sure that requests to quay will be done.

mmorhun, line 111: Does it make sense to have a factory?

controllers/component_image_controller.go
```diff
@@ -458,12 +461,23 @@ func generateRepositoryName(component *appstudioredhatcomv1alpha1.Component) str
 	return component.Namespace + "/" + component.Spec.Application + "/" + component.Name
 }
 
-func (r *ComponentReconciler) generateImageRepository(ctx context.Context, component *appstudioredhatcomv1alpha1.Component, opts *GenerateRepositoryOpts) (*quay.Repository, *quay.RobotAccount, *quay.RobotAccount, error) {
+func (r *ComponentReconciler) generateImageRepository(
+	ctx context.Context,
+	quayClient quay.QuayService,
+	organization string,
```
mmorhun, line 467: It doesn't look right to have it in the method params and the struct itself.

main.go
```diff
@@ -43,6 +45,11 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
+const (
+	tokenPath string = "/workspace/quaytoken"
+	orgPath   string = "/workspace/organization"
```
mmorhun, line 50: I'd explicitly add word "quay" to the names.

mmorhun (commented):
> Also, with this approach you are going to have a mismatch between your configuration (what you have in the secret) and the data the controller is using.
It was a bit confusing to me, when I updated the secret after I found a typo in the organization name.

@filariow RHTAP is not designed to be able to switch quay organization on the go. While you can still do it technically, it won't work smoothly, because a migration is needed for existing components. Only quay token rotation makes sense to me.

controllers/component_image_controller.go
filariow: ok, I'm going to instantiate the client just before use

controllers/component_image_controller.go
filariow: as we are removing Organization read at runtime, we are going to have a read at startup for the Organization and a read if the client is needed.

controllers/component_image_controller.go
filariow: Removed the organization. To avoid duplicate instantiation of the quayClient, the quayClient is passed from Reconcile.

mmorhun (commented):
@filariow thank you for the update. However, current implementation is somewhat twisted with new types, lazy and not lazy clients. That makes code more complicated and harder to understand and maintain.
Could we leave everything as it was, except adding a new method to the reconciler that just rereads the config. So, instead of using all these clients and passing additional args, we just add this new method and call it in two places. What do you think?

filariow:
> @filariow thank you for the update. However, current implementation is somewhat twisted with new types, lazy and not lazy clients. That makes code more complicated and harder to understand and maintain. Could we leave everything as it was, except adding a new method to the reconciler that just rereads the config. So, instead of using all these clients and passing additional args, we just add this new method and call it in two places. What do you think? P.S. we probably need to pass a function to the reconciler as an argument to avoid filesystem reading in tests.

@mmorhun thanks for the review; I update the PR removing the Lazy and updating the `BuildQuayClientFunc` to take the logger from outside, so it won't always use the setupLog, but the one the reconciler is configuring for the reconcile's context. I agree it's much cleaner this way and the Lazy approach is not really needed in this case.
Btw, it's not completely clear to me the approach you are suggesting. It seems to me, the new method in the reconciler would do the same thing as the one we should provide as a function for avoiding filesystem reading in tests. Also, this externally provided function to me is the `BuildQuayClient` one. May you clarify your suggestion?

mmorhun:
@filariow sorry if I was not clear enough. The purpose of that method is the same as in `BuildQuayClient`.

mmorhun (approved):
Tested, it updates the token after some period of time. The delay happens because of asynchronous processes in the cluster itself.

mmorhun:
@filariow please fix linter errors

Michkov (approved):

filariow:
> @filariow please fix linter errors

Done. I just ignored the errors as they are not applying. For the `Potential file inclusion via variable` an alternative is to sanitize the input path with `cp := filepath.Clean(path)`, but as the input is a `const` fixed by design, I thought it is not necessary to clean it up.

mmorhun:
@filariow there is one more lint error about io util.

mmorhun:
@filariow thank you for the contribution! :+1: 
I'll be in PTO starting from the next week, so kindly asking @Michkov to merge this PR when ready.

mmorhun (approved):

PR #38, opened 2023-07-06
purge old pipelinerun defs

sbose78:


sbose78:
as part of https://github.com/redhat-appstudio/infra-deployments/pull/2061

PR #42, opened 2023-07-10
test: improve coverage of pkg/quay

tnevrlka:
pkg/quay currently has some completely uncovered functions, making the coverage 33 %
This PR adds tests in order to increase coverage

### DIFF
--- a/pkg/quay/quay_test.go
+++ b/pkg/quay/quay_test.go
@@ -26,6 +26,12 @@ import (
 	"github.com/h2non/gock"
 )
 
+const (
+	org       = "test_org"
+	repo      = "test_repo"
+	robotName = "robot_name"
+)
+
 func TestQuayClient_CreateRepository(t *testing.T) {
 	defer gock.Off()
 
@@ -293,9 +299,6 @@ func TestQuayClient_handleRobotName(t *testing.T) {
 func TestQuayClient_GetTagsFromPage(t *testing.T) {
 	defer gock.Off()
 
-	org := "test_org"
-	repo := "test_repo"
-
 	testCases := []struct {
 		name          string
 		pages         int
@@ -360,9 +363,6 @@ func TestQuayClient_GetTagsFromPage(t *testing.T) {
 func TestQuayClient_DeleteTag(t *testing.T) {
 	defer gock.Off()
 
-	org := "test_org"
-	repo := "test_repo"
-
 	testCases := []struct {
 		name       string
 		tag        string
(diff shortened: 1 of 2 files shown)


pkg/quay/quay_test.go
```diff
@@ -425,3 +425,180 @@ func TestQuayClient_DeleteTag(t *testing.T) {
...
+
+	// Get existing robot account
+	gock.New("https://quay.io/api/v1").
+		MatchHeader("Content-Type", "application/json").
+		MatchHeader("Authorization", "Bearer authtoken").
+		Get(fmt.Sprintf("organization/%s/robots/%s", org, robotName)).
+		Reply(200).
+		JSON(robot)
+
+	if received_robot, err := quayClient.GetRobotAccount(org, robotName); err != nil {
```
tisutisu, line 555: nit: Is it better to avoid snake case and stick to camel case `received_robot`

pkg/quay/quay_test.go
```diff
@@ -425,3 +425,180 @@ func TestQuayClient_DeleteTag(t *testing.T) {
...
+	} else if *received_robot != robot {
+		t.Error("received robot is not the same as original")
+	}
+
+	// Try to get non-existing robot account
+	gock.New("https://quay.io/api/v1").
+		MatchHeader("Content-Type", "application/json").
+		MatchHeader("Authorization", "Bearer authtoken").
+		Get(fmt.Sprintf("organization/%s/robots/%s", org, robotName)).
+		Reply(404)
```
tkdchen, line 566: I'm not familiar with gock. If `.JSON(...)` is not called, what fake value will be used by gock? Is it same as the call `.JSON(map[string]string{})`?

tkdchen (commented):
Nice! Can you please cover more code path, for example the error when server does not respond data in an encoded JSON string ?

Another suggestion, can you reorganize the test structure as https://github.com/redhat-appstudio/build-service/blob/main/controllers/component_build_controller_unit_test.go#L42? So that in a test, we can clearly define fake input (simulate Quay.io response) and the expected return value for each test case.

tnevrlka:
Is it possible that `GetRobotAccount` is behaving unexpectedly? Could someone please double check?
I believe it does not return error when Unauthorized

pkg/quay/quay_test.go
tnevrlka: Thanks, that's what happens when I take a week long break from Go :)

pkg/quay/quay_test.go
tnevrlka: Thank you. I rewrote the tests and I believe it solved the uncertainty about `.JSON(...)`

Michkov (approved):

PR #45, opened 2023-07-14
Updated dockerfile to use redhat based images

tisutisu:
This PR addresses story: https://issues.redhat.com/browse/STONEBLD-1498

PR #47, opened 2023-07-17
STONEBLD-1498 Update base images to rhel based

Michkov:
Updating Dockerfile to use RH based base images. Removed test part since there is separate github workflow.


Michkov:
replacement of #45

tisutisu (approved):
LGTM

PR #48, opened 2023-07-17
ci: fail codecov workflow if error happens

tnevrlka:
Codecov upload on push to main is currently encountering an error, but the GitHub workflow does not fail like it does on pull request. 

This PR will make the worfklow fail if it encounters an error, making it easier to see the error


tnevrlka:
The error seems to be gone. But the need for the CI to fail is still there, so we can easily see if another error happens

PR #53, opened 2023-08-16
test: Add test cases to pkg/quay

tkdchen:
* Rewrite part of the test functions in order to add test cases conveniently.
* handleRobotName is refactored slightly. 1) Once the input robotName is matched, no need the check the number of parts separated by +. 2) The regexp is hardcoded and not an input from somewhere else, and the tests ensure that the regexp is correct and can be compiled by the library. So, no need to handle the err of an invalid regexp.
* Unify the type of response data passed to gock JSON func so that string value can be passed as well.
* Assertions are written with gotest.tools assert functions. One of the benefits is the error can be easily checked if it is nil and if a specific err is returned by specifying part of the error message.

### DIFF
--- a/pkg/quay/quay.go
+++ b/pkg/quay/quay.go
@@ -509,20 +509,13 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 // e.g. `org+robot` will be changed to `robot`, `robot` will stay `robot`
 func handleRobotName(robotName string) (string, error) {
 	// Regexp from quay api `^([a-z0-9]+(?:[._-][a-z0-9]+)*)$` with one plus sign in the middle allowed (representing longname)
-	r, err := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
+	r := regexp.MustCompile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
 	robotName = strings.TrimSpace(robotName)
-	if err != nil {
-		return "", fmt.Errorf("failed to compile regex, error: %s", err)
-	}
 	if !r.MatchString(robotName) {
 		return "", fmt.Errorf("robot name is invalid, must match `^([a-z0-9]+(?:[._-][a-z0-9]+)*)$` (one plus sign in the middle is also allowed)")
 	}
 	if strings.Contains(robotName, "+") {
-		parts := strings.Split(robotName, "+")
-		if len(parts) != 2 {
-			return "", fmt.Errorf("robotName could not be split into two parts, expected len 2, got len %d", len(parts))
-		}
-		robotName = parts[1]
+		robotName = strings.Split(robotName, "+")[1]
 	}
 	return robotName, nil
 }
--- a/pkg/quay/quay_test.go
+++ b/pkg/quay/quay_test.go
@@ -20,8 +20,9 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
-	"strings"
 	"testing"
+
+	"gotest.tools/v3/assert"
 
 	"github.com/h2non/gock"
 )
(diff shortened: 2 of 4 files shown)


pkg/quay/quay.go
```diff
@@ -120,7 +120,8 @@ func (c *QuayClient) DoesRepositoryExist(organization, imageRepository string) (
 
 	if res.StatusCode == 404 {
 		return false, fmt.Errorf("repository %s does not exist in %s organization", imageRepository, organization)
-	} else if res.StatusCode == 200 {
+	}
+	if res.StatusCode == 200 {
```
mkosiarc, line 124: why this change?

pkg/quay/quay.go
```diff
@@ -509,20 +510,13 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 // e.g. `org+robot` will be changed to `robot`, `robot` will stay `robot`
 func handleRobotName(robotName string) (string, error) {
 	// Regexp from quay api `^([a-z0-9]+(?:[._-][a-z0-9]+)*)$` with one plus sign in the middle allowed (representing longname)
-	r, err := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
+	r, _ := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
 	robotName = strings.TrimSpace(robotName)
-	if err != nil {
```
mkosiarc, line 514: we now it compiles now. Is there not a chance that we would update this in the future?

pkg/quay/quay.go
tkdchen: Restored.

mkosiarc (approved):

pkg/quay/quay_test.go
```diff
@@ -20,199 +20,461 @@ import (
...
+
 	"github.com/h2non/gock"
 )
 
 const (
 	org       = "test_org"
 	repo      = "test_repo"
 	robotName = "robot_name"
+
+	testRepoName        = "test-repo-using-api"
```
tnevrlka, line 35: Is there a need for `testRepoName` instead of using `repo`?

pkg/quay/quay_test.go
```diff
@@ -20,199 +20,461 @@ import (
...
+
+			receivedRepos, err := quayClient.GetAllRepositories("test_org")
+
+			if tc.expectedErr == "" {
+				if err != nil {
+					t.Errorf("Error getting all repositories, Expected nil, got %v", err)
+				}
+				if len(receivedRepos) != 3 {
+					t.Errorf("Possible pagination error, expected 3 repos, got %d repos", len(receivedRepos))
+				}
```
tnevrlka, line 413: `assert.NilError` and `assert.Condition` could be used here, couldn't they?

pkg/quay/quay_test.go
```diff
@@ -337,75 +626,92 @@ func TestQuayClient_GetTagsFromPage(t *testing.T) {
...
+				tags, hasAdditional, err := quayClient.GetTagsFromPage(org, repo, page)
+				if tc.expectedErr == "" {
+					assert.NilError(t, err)
+
+					if !reflect.DeepEqual(mockTags, tags) {
+						t.Errorf("tags are not the same, expected `%v`, got `%v`", mockTags, tags)
+					}
+					if hasAdditional != tc.hasAdditional[page-1] {
+						t.Errorf("hasAdditional is not the same, expected `%t`, got `%t`", tc.hasAdditional[page-1], hasAdditional)
+					}
```
tnevrlka, line 654: `assert.DeepEqual` and `assert.Condition` might work here?

tnevrlka (approved):
Thank you, very nice

pkg/quay/quay_test.go
```diff
@@ -20,199 +20,461 @@ import (
...
-		MatchHeader("Authorization", "Bearer authtoken").
-		Post("/repository").
-		Reply(200).JSON(map[string]string{
-		"description": "description",
-		"namespace":   "redhat-appstudio-user",
-		"name":        "test-repo-using-api",
-	})
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			defer gock.Off()
```
Michkov, line 108: defer is called at the end of the function, so it can be just called once at beginning. Am I right @mmorhun

pkg/quay/quay_test.go
tkdchen: Good catch. This was refactored from corresponding test. Updated to use `repo`.

pkg/quay/quay_test.go
tkdchen: Updated. PTAL.

pkg/quay/quay_test.go
tkdchen: Calling defer in the anonymous function intentionally due to some test function failed without it. But, I forget which one :sweat_smile:

pkg/quay/quay.go
```diff
@@ -559,20 +559,13 @@ func (c *QuayClient) GetAllRobotAccounts(organization string) ([]RobotAccount, e
 // e.g. `org+robot` will be changed to `robot`, `robot` will stay `robot`
 func handleRobotName(robotName string) (string, error) {
 	// Regexp from quay api `^([a-z0-9]+(?:[._-][a-z0-9]+)*)$` with one plus sign in the middle allowed (representing longname)
-	r, err := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
+	r, _ := regexp.Compile(`^[a-z0-9]+(?:[._-][a-z0-9]+)*(?:\+[a-z0-9]+(?:[._-][a-z0-9]+)*)?$`)
```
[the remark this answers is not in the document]
tkdchen: Updated.

pkg/quay/quay_test.go
tkdchen: Updated.

PR #65, opened 2023-09-20
Revert readOnlyRootFilesystem for image_pruner

hugares:
The image-pruner container needs write on rootFS as it is installing pip packages. Put back write access until image-pruner is changed to not require this access.


mmorhun (approved):

mmorhun:
Merging this as a temporary workaround.
We'll work on proper fix.

PR #67, opened 2023-09-26
Use Remote Secret for storing image repository push token

mmorhun:
This PRs makes push token stored in a RemoteSecret instead of just Secret. Such approach gives the following advantages:
* Ability to propagate the token to BYOC and remote clusters
* Makes Image Controller backup / restore friendly (avoiding backing up secrets directly)
* Simplifies self healing implementation
* Allows to have the same approach for push and pull tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ status:
   credentials:
     generationTimestamp: "2023-08-23T14:56:41Z"
     push-robot-account: test_ns_imagerepository_sample_101e4e2b63
-    push-secret: test-ns-imagerepository-sample-101e4e2b63
+    push-remote-secret: imagerepository-sample-image-push
+    push-secret: imagerepository-sample-image-push
   image:
     url: quay.io/my-org/test-ns/imagerepository-sample
     visibility: public
@@ -63,6 +64,7 @@ status:
 ```
 where:
   - `push-robot-account` is the name of quay robot account in the configured quay organization with write premissions to the repository.
+  - `push-remote-secret` is an instance of `RemoteSecret` that manages the `Secret` specified in `push-secret`.
   - `push-secret` is a `Secret` of dockerconfigjson type that contains image repository push robot account token with write permissions.
 
 ### User defined image repository name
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -127,6 +129,8 @@ After any successful operation, `status.message` is cleared.
 
 ## AppStudio Component image repository
 
+### Image repository for Component builds specifics
```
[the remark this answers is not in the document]
mmorhun: I meant some Component only things, but probably better to remove that word at all to avoid confusion.

api/v1alpha1/imagerepository_types.go
```diff
@@ -117,6 +117,13 @@ type CredentialsStatus struct {
 	// PullRobotAccountName is present only if ImageRepository has labels that connect it to Application and Component.
 	// Holds name of the quay robot account with real (pull only) permissions from the generated repository.
 	PullRobotAccountName string `json:"pull-robot-account,omitempty"`
+
+	// PushRemoteSecretName holds name of RemoteSecret object that manages push Secret and its linking to pipeline Service Account.
```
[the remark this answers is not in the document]
mmorhun: Yes, I did. Will update the comment.

tisutisu (approved):
LGTM

controllers/imagerepository_controller.go
```diff
@@ -500,13 +491,16 @@ func (r *ImageRepositoryReconciler) EnsureRemotePullSecret(ctx context.Context,
 			return err
 		}
 
+		serviceAccountName := buildPipelineServiceAccountName
+		if isPull {
```
rcerven, line 495: why are you using default SA for pull secret?

controllers/imagerepository_controller.go
```diff
@@ -518,14 +512,22 @@ func (r *ImageRepositoryReconciler) EnsureRemotePullSecret(ctx context.Context,
...
+									Name: serviceAccountName,
 								},
 							},
 						},
 					},
 				},
 			},
 		}
+
+		if isPull {
```
rcerven, line 524: also don't understand why this differs for pull & push secret

controllers/imagerepository_controller.go
mmorhun: Because there is no build pipeline SA in that namespaces and also no guarantee that any other SA exists.

controllers/imagerepository_controller.go
mmorhun: Because push tokne must be in current (build) namespace only (at least for current requirements), but pull token in all environments of the application.

PR #76, opened 2023-11-09
Update dependencies

mmorhun:


rcerven (approved):

main.go
```diff
@@ -90,12 +91,20 @@ func main() {
...
+	}
+	metricsOpts := server.Options{
+		BindAddress: metricsAddr,
+	}
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Client:                 clientOpts,
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
```
chmeliik, line 96: We don't need to set the port anymore?

mmorhun: It's scaffolded deprecated configuration for webhook port which is not used in Image Controller.

chmeliik (approved):
LGTM :+1:

PR #80, opened 2023-11-23
feat: add sonar cloud properties for better control

sawood14012:
Adding Sonar Cloud Properties File for better control

https://issues.redhat.com/browse/STONEBLD-1940


chmeliik (commented):
We don't need the other properties like we have in build-service? https://github.com/redhat-appstudio/build-service/blob/main/.sonarcloud.properties

.sonarcloud.properties
```diff
@@ -0,0 +1,9 @@
+# Path to sources
+sonar.sources=.
+sonar.exclusions=**/*test*, **/vendor/**, **/zz_generated.*, **/*.yaml,
+
+# Source encoding
+sonar.sourceEncoding=UTF-8
+
+# Exclusions for copy-paste detection
+sonar.cpd.exclusions=**/*test*
```
chmeliik, line 9: Could you add the missing newline?

sawood14012:
> We don't need the other properties like we have in build-service? https://github.com/redhat-appstudio/build-service/blob/main/.sonarcloud.properties

I had added those just to identify those are not required i am planning to remove them from build service also

chmeliik (approved):
LGTM :+1:

PR #84, opened 2023-12-21
reporting request times for each action to prometheus for SLO/SLI

rcerven:
STONEBLD-1775


mkosiarc (approved):

controllers/component_image_controller.go
```diff
@@ -108,7 +113,12 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("error reading component: %w", err)
 	}
 
+	componentIdForMetrics := fmt.Sprintf("%s=%s", component.Name, component.Namespace)
```
mmorhun, line 116: I'd create a separate function `getComponentIdForMetrics(component)`. This way, we won't need to change signature of `reportError` method.

controllers/imagerepository_controller.go
```diff
@@ -60,13 +70,52 @@ type ImageRepositoryReconciler struct {
...
+	if err := initMetrics(); err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&imagerepositoryv1alpha1.ImageRepository{}).
 		Complete(r)
 }
 
+func updateMetricsTimes(componentIdForMetrics string, reconcileStartTime time.Time) {
```
mmorhun, line 111: This function is about `set` not `update`.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/h2non/gock v1.2.0
 	github.com/onsi/ginkgo/v2 v2.13.0
 	github.com/onsi/gomega v1.29.0
+	github.com/prometheus/client_golang v1.17.0
```
mmorhun, line 10: I guess `go.sum` should be updated as well.

controllers/imagerepository_controller.go
```diff
@@ -109,6 +164,9 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	if imageRepository.Status.State == imagerepositoryv1alpha1.ImageRepositoryStateFailed {
+		// remove component from metrics map
```
mmorhun, line 167: Do we need failures metric?

controllers/imagerepository_controller.go
```diff
@@ -161,6 +220,15 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 	}
 
+	// we are adding to map only for new provision, not for some partial actions,
+	// so report time only if time was recorded
+	_, timeRecorded := repositoryTimesForMetrics[componentIdForMetrics]
+	if timeRecorded {
+		imageRepositoryProvisionTimeMetric.Observe(time.Since(repositoryTimesForMetrics[componentIdForMetrics]).Seconds())
```
mmorhun, line 227: This code a bit twisted.
1. The ignored flag can be used in the if condition
2. `timeRecorded` is already here, no need to query map again.

controllers/imagerepository_controller.go
rcerven: that isn't for failures metric, but just for cleaning that map, in case full reconcile didn't finish but failed

go.mod
rcerven: it wasn't added but just moved from "indirect" up to direct, so go.sum has it

controllers/imagerepository_controller.go
rcerven: ok I will not ignore value and then not query it again

controllers/imagerepository_controller.go
rcerven: renamed

controllers/component_image_controller.go
rcerven: done

controllers/imagerepository_controller.go
```diff
@@ -161,6 +220,15 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	}
 
+	// we are adding to map only for new provision, not for some partial actions,
+	// so report time only if time was recorded
+	provisionTime, timeRecorded := repositoryTimesForMetrics[componentIdForMetrics]
+	if timeRecorded {
+		imageRepositoryProvisionTimeMetric.Observe(time.Since(provisionTime).Seconds())
+	}
+	// remove component from metrics map
+	delete(repositoryTimesForMetrics, componentIdForMetrics)
```
mmorhun, line 230: We can do the deletion only if the time was recorded.

controllers/imagerepository_controller.go
```diff
@@ -60,13 +70,52 @@ type ImageRepositoryReconciler struct {
...
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&imagerepositoryv1alpha1.ImageRepository{}).
 		Complete(r)
 }
 
+func setMetricsTime(componentIdForMetrics string, reconcileStartTime time.Time) {
+	_, timeRecorded := repositoryTimesForMetrics[componentIdForMetrics]
+
```
mmorhun, line 113: Looks like an extra line.

rcerven: ok

controllers/imagerepository_controller.go
rcerven: time will be always recorded at that point

controllers/imagerepository_controller.go
rcerven: added

tkdchen (commented):
Function `initMetrics` is called twice inside both reconciler `SetupWithManager` functions. I assume the duplicated calls should not cause any potential problem due to the underlying package handles that. Even though, how about to avoid duplicated call explicitly?

Metrics related code are mainly implemented inside imagerepository_controller.go file and shared with ComponentReconciler. My suggestion is to make a separate go file under package `controllers` to include those shared code related to metrics.

controllers/imagerepository_controller.go
```diff
@@ -90,7 +151,12 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	componentIdForMetrics := fmt.Sprintf("%s=%s", imageRepository.Labels[ComponentNameLabelName], imageRepository.Namespace)
```
mmorhun, line 154: ImageRepository doesn't have to have `ComponentNameLabelName` label. For example, JBS uses it without the label.
I think we have to have identifier not related to Component, but to instance of ImageRepository itself.

controllers/imagerepository_controller.go
mmorhun: Then why do we need that `if` above?

controllers/component_image_controller.go
```diff
@@ -558,3 +579,7 @@ func generateDockerconfigSecretData(quayImageURL string, robotAccount *quay.Robo
 		quayImageURL, base64.StdEncoding.EncodeToString([]byte(authString)))
 	return secretData
 }
+
+func getComponentIdForMetrics(component *appstudioredhatcomv1alpha1.Component) string {
+	return component.Name + "=" + component.Namespace
+}
```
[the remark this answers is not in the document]
rcerven: well that uses different thing and not component

controllers/imagerepository_controller.go
rcerven: I mean in this case I guess I can just move that delete inside if

rcerven:
> Function `initMetrics` is called twice inside both reconciler `SetupWithManager` functions. I assume the duplicated calls should not cause any potential problem due to the underlying package handles that. Even though, how about to avoid duplicated call explicitly?
> 
> Metrics related code are mainly implemented inside imagerepository_controller.go file and shared with ComponentReconciler. My suggestion is to make a separate go file under package `controllers` to include those shared code related to metrics.

did you see the first thing I do in initMetrics??

	// don't register it if it was already registered by another controller
	if imageRepositoryProvisionTimeMetric != nil {
		return nil
	}

that ensures it doesn't run multiple times

but I have to call it  in both controllers because 1 of them will be first to initialize it

controllers/imagerepository_controller.go
rcerven: will use name + namespace then

mmorhun (approved):

PR #88, opened 2024-01-05
Replace requests with urllib in prune_images script

tkdchen:
[STONEBLD-1785](https://issues.redhat.com//browse/STONEBLD-1785)

This replacement ensures no write operation to the rootFS happens during the job execution.

Major changes:

* Part of the prune_images script is rewritten with urllib instead of requests.
* Test code is also rewritten so that prune_images script uses all builtin libraries for execution and development. No write operation is required for running the script during the cronjob execution.
* Additional tests are added.

Run tests:

```bash
cd config/registry_image_pruner/image_pruner/
python3.9 test_prune_images.py
```

In addition, two small refactors are made and a GitHub workflow action is added to run the tests in CI.


config/registry_image_pruner/cronjob.yaml
```diff
@@ -29,9 +29,7 @@ spec:
             command:
               - /bin/bash
               - '-c'
-              - >-
-                pip install -r /image-pruner/requirements.txt && python
-                /image-pruner/prune_images.py --namespace $(NAMESPACE)
+              - python /image-pruner/prune_images.py --namespace $(NAMESPACE)
```
chmeliik, line 32: Hmm, how does this work? `$(NAMESPACE)` doesn't look like the correct syntax for a bash env var :thinking:

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -13,41 +19,87 @@
...
-                    LOGGER.info("Removing image %s from %s", image, repository)
-                    delete_image_endpoint = (
-                        f"{QUAY_API_URL}/repository/{repository}/tag/{image}"
-                    )
-                    response = session.delete(delete_image_endpoint)
-                    response.raise_for_status()
+                    LOGGER.info("Removing image %s from %s/%s", image, namespace, name)
+                    delete_image_repo(quay_token, namespace, name, image)
+        else:
+            LOGGER.debug("%s is not an image with suffix .att or .sbom", image)
```
chmeliik, line 62: Is this log message useful? It seems like it would make the output too verbose

config/registry_image_pruner/image_pruner/tests/requirements.txt
```diff
@@ -1,2 +0,0 @@
-pytest
-flexmock
```
chmeliik, line 2: It was probably not necessary to rework tests, no? These don't get installed in the job, only on the developer's machine.

Reducing the tests changes would make it possible to see what tests were added in this commit, currently it's hard to review

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -13,41 +19,87 @@
...
+            json_data = json.loads(resp.read())
 
-        repository_endpoint = f"{QUAY_API_URL}/repository/{repository}"
-        response = session.get(repository_endpoint)
-        response.raise_for_status()
-        repository_json = response.json()
+        repos = json_data.get("repositories", [])
+        if not repos:
+            LOGGER.debug("No image repository is found.")
+            break
```
chmeliik, line 96: Do we want to `break` when there are no repositories? It looks like the script previously didn't do that

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -47,11 +47,11 @@ def delete_image_repo(quay_token: str, namespace: str, name: str, tag: str) -> N
 
 def remove_images(images: Dict[str, Any], quay_token: str, namespace: str, name: str, dry_run: bool = False) -> None:
     image_digests = [image["manifest_digest"] for image in images.values()]
+    image_regex = re.compile(r"^sha256-([0-9a-f]+)(\.sbom|\.att)$", re.IGNORECASE)
     for image in images:
         # attestation or sbom image
-        if image.endswith(".att") or image.endswith(".sbom"):
-            sha = re.search("sha256-(.*)(.sbom|.att)", image).group(1)
-            if f"sha256:{sha}" not in image_digests:
+        if (match := image_regex.match(image)) is not None:
```
chmeliik, line 54: Is `image` just a tag? If it's a full image name, it won't match something that starts with `sha256-...`

mkosiarc:
I know you already did the work, but I would prefer if we just had an image with the requests library in our new dockerfiles repos

config/registry_image_pruner/image_pruner/prune_images.py
tisutisu: ran the script on my `susdas_quay_org` quay repo, the script is not removing `.sbom` or `.att` tags, as it needs to be, below is an example images dictionary which is passed to the `remove_images` method 
```
images = {'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.att': {'name': 'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.att', 'size': 89162, 'last_modified': 'Mon, 21 Aug 2023 07:09:56 -0000', 'manifest_digest': 'sha256:52508a6193fdfa67f125f47a632ee6420f34d729776752890a5c6e00dbad6725'}, 
          'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.sig': {'name': 'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.sig', 'size': 307, 'last_modified': 'Mon, 21 Aug 2023 07:08:33 -0000', 'manifest_digest': 'sha256:0e3c6b8cdde71673336dc988425dbc0283445ae72878a675300399b31599e525'}, 
          'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.sbom': {'name': 'sha256-876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d.sbom', 'size': 114547, 'last_modified': 'Mon, 21 Aug 2023 07:08:29 -0000', 'manifest_digest': 'sha256:7cd13d1b83d45e7f7a9c526b2264c224c858f03f3dd27eaa4980fc44a0c50709'}, 
          'build-8ca40-1692601660': {'name': 'build-8ca40-1692601660', 'size': 16091103, 'last_modified': 'Mon, 21 Aug 2023 07:08:27 -0000', 'manifest_digest': 'sha256:876b86c27ac4132a1660f7325fd2350a6b9f792ce30a87cfa626753b7ca98a3d'}
}
```

config/registry_image_pruner/image_pruner/tests/requirements.txt
tkdchen: The rework covers all previous existing tests.

With the replacement, the existing tests have to be reworked by mocking calls to urllib properly and it looks no difference too much with the current implementation except using different mock framework and test runner. And for a Go project, I would like to keep it simple enough with builtin libraries for a single Python script.

chmeliik: ok

config/registry_image_pruner/image_pruner/prune_images.py
chmeliik: > `image` is tag.

Could you maybe rename the variable to make it clearer?

tisutisu: @tkdchen then does the `prune_images.py` script doing what it is supposed to (cleaning up of .att and .sbom image tags) ?

mkosiarc: I would not say that `image` is a tag. It's separate image even though it is under the "tags" section in quay (this is just the rhtap/quay integration design). The .att .sbom images are associated with the build image, but they are not  just tags for that image and they should be removed

mkosiarc: It would be nice to have a unit test specifically for remove images, where various scenarios would be tested like the old unit tests had so we can verify it works as expected

mkosiarc: I can now confirm that the script does what it is supposed to
```
(venv) [mkosiarc@fedora image_pruner]$ git branch
  image-pruner
  main
* tkdchen-STONEBLD-1785-pruner
# nothing to do, build image present
(venv) [mkosiarc@fedora image_pruner]$ QUAY_TOKEN=<TOKEN> python prune_images.py --namespace mkosiarc_rhtap
2024-01-09 14:57:33,957 - INFO - Processing repository 0: mkosiarc_rhtap/test-pvc/test-application/sample-app-newer-python-image
# I removed the build image, the script should remove the .att and .sbom images
(venv) [mkosiarc@fedora image_pruner]$ QUAY_TOKEN=<TOKEN> python prune_images.py --namespace mkosiarc_rhtap
2024-01-09 14:57:49,377 - INFO - Processing repository 0: mkosiarc_rhtap/test-pvc/test-application/sample-app-newer-python-image
2024-01-09 14:57:50,224 - INFO - Removing image sha256-f685cce54b2d4491a4a8b7a3bc6db437190a28795684ebe8271c8d3ad6cf9120.att from mkosiarc_rhtap/test-pvc/test-application/sample-app-newer-python-image
2024-01-09 14:57:51,308 - INFO - Removing image sha256-f685cce54b2d4491a4a8b7a3bc6db437190a28795684ebe8271c8d3ad6cf9120.sbom from mkosiarc_rhtap/test-pvc/test-application/sample-app-newer-python-image
```

config/registry_image_pruner/image_pruner/prune_images.py
mkosiarc: I previously chose that approach because it felt more pythonic to me
as the first line of the function 
```
for repo in repos:
```
would just ignore empty repos. This way we have to have a condition and a break, but if the images can really be `None` then this approach make sense

config/registry_image_pruner/cronjob.yaml
mkosiarc: that's a kubernetes thingy https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

chmeliik: nice, TIL

config/registry_image_pruner/image_pruner/prune_images.py
tkdchen: Hi @mkosiarc, thanks for sharing your thought. Regarding the unit tests, `remove_images` has been covered 100% already.

tkdchen: > > `image` is tag.
> 
> Could you maybe rename the variable to make it clearer?

Renamed.

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -1,53 +1,104 @@
...
     image_digests = [image["manifest_digest"] for image in images.values()]
-    for image in images:
-        # attribute or sbom image
-        if image.endswith(".att") or image.endswith(".sbom"):
-            sha = re.search("sha256-(.*)(.sbom|.att)", image).group(1)
-            if f"sha256:{sha}" not in image_digests:
+    image_regex = re.compile(r"^sha256-([0-9a-f]+)(\.sbom|\.att)$")
+    for tag in images:
+        # attestation or sbom image
+        if (match := image_regex.match(tag)) is not None:
```
chmeliik, line 54: nitpick: with the rename of `image` -> `tag`, the regex should probably be called `tag_regex`

chmeliik (approved):
LGTM :+1:

config/registry_image_pruner/image_pruner/prune_images.py
mkosiarc: since we decided to rename, when iterating 
```
for tag in images
```
I think we are iterating through all the tags for certain image (in quay) so should be something like for tag in tags
also, then we could change the name of the method as well.

So I feel like we should not mix those names if they refer to the same thing

config/registry_image_pruner/image_pruner/prune_images.py
mkosiarc: Thanks chen for adding those tests, I appreciate it.

config/registry_image_pruner/image_pruner/prune_images.py
tkdchen: @mkosiarc PTAL. More replacements are made.

mkosiarc (approved):

PR #99, opened 2024-03-26
Pin to a non-breaking version of setup-envtest

chmeliik:
https://github.com/kubernetes-sigs/controller-runtime/issues/2720

The minimum required Go version for setup-envtes was bumped to 1.22, breaking us. Pin to release-0.16, which doesn't use 1.22 yet. It uses 1.21, which is still higher than what we use, but back then Go didn't enforce the version requirement yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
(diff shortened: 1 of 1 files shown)


chmeliik:
Should fix the failing tests https://github.com/redhat-appstudio/image-controller/actions/runs/8439729454/job/23115019425?pr=95

rcerven (approved):

tkdchen (approved):

mmorhun (approved):

chmeliik:
Don't understand why EC failed, all those tasks are there
image

chmeliik:
Maybe it's failing because the task bundles are outdated? I'll combine this PR with #95 again

chmeliik:
Merged in #95

PR #103, opened 2024-04-04
Dockerfile: update the ubi8-minimal base image

chmeliik:


chmeliik:
Does not address the EC failure, closing

PR #105, opened 2024-04-12
Image pruner should get all tags and not only first 500

rcerven:


config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -24,24 +24,42 @@
...
-        if ex.status != 404:
-            raise(ex)
+        tags = json_data.get("tags", [])
+        all_tags.extend(tags)
+
+        if not tags:
+            LOGGER.debug("No tags found.")
+            break
+
+        page = json_data.get("page", None)
```
lcarva, line 54: Maybe `0` is a more meaningful default value?

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -24,24 +24,42 @@
...
+        additional = json_data.get("has_additional", False)
+
+        if additional:
+            next_page = page + 1
         else:
-            return {}
+            break
```
lcarva, line 60: Why break here?

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -24,24 +24,42 @@
...
-            raise(ex)
+        tags = json_data.get("tags", [])
+        all_tags.extend(tags)
+
+        if not tags:
+            LOGGER.debug("No tags found.")
+            break
+
+        page = json_data.get("page", None)
+        additional = json_data.get("has_additional", False)
```
mmorhun, line 55: It could be a condition for the while, so you can avoid `break` below.

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -61,34 +79,33 @@ def delete_image_tag(quay_token: str, namespace: str, name: str, tag: str) -> No
...
-        if (match := tag_regex.match(tag)) is not None:
+        if (match := tag_regex.match(tag["name"])) is not None:
             if f"sha256:{match.group(1)}" not in image_digests:
                 if dry_run:
-                    LOGGER.info("Image %s from %s/%s should be removed", tag, namespace, name)
+                    LOGGER.info("Image %s from %s/%s should be removed", tag["name"], namespace, name)
                 else:
-                    LOGGER.info("Removing image %s from %s/%s", tag, namespace, name)
-                    delete_image_tag(quay_token, namespace, name, tag)
+                    LOGGER.info("Removing image %s from %s/%s", tag["name"], namespace, name)
```
mmorhun, line 92: Is it really image or tag?

config/registry_image_pruner/image_pruner/prune_images.py
rcerven: it doesn't really matter tbh

config/registry_image_pruner/image_pruner/prune_images.py
rcerven: because we don't have any new additional pages

config/registry_image_pruner/image_pruner/prune_images.py
rcerven: tbh this seems to be more readable, just to safe 2 lines of else & break

config/registry_image_pruner/image_pruner/prune_images.py
rcerven: well :D actually we are removing here those additional artifacts

but we are printing the whole artifact name, so it is clear

config/registry_image_pruner/image_pruner/prune_images.py
lcarva: ack. Misread the code earlier.

mmorhun (approved):
LGTM, but please use tag instead of image in teh message as image can be confused with whole repository.

PR #113, opened 2024-05-20
Remove deprecated source image tags

tkdchen:
[STONEBLD-2122](https://issues.redhat.com//browse/STONEBLD-2122)

These kinds deprecated tags are removed:

* the associated binary image does not exist.
* the coexistent one with the new tag.

### DIFF
--- a/config/registry_image_pruner/image_pruner/prune_images.py
+++ b/config/registry_image_pruner/image_pruner/prune_images.py
@@ -81,6 +81,7 @@ def delete_image_tag(quay_token: str, namespace: str, name: str, tag: str) -> No
 
 def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, name: str, dry_run: bool = False) -> None:
     image_digests = [image["manifest_digest"] for image in tags]
+    tags_map = {tag_info["name"]: tag_info for tag_info in tags}
     tag_regex = re.compile(r"^sha256-([0-9a-f]+)(\.sbom|\.att|\.src|\.sig)$")
     for tag in tags:
         # attestation or sbom image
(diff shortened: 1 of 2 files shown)


config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -91,8 +91,11 @@ def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, nam
                 else:
                     LOGGER.info("Removing tag %s from %s/%s", tag["name"], namespace, name)
                     delete_image_tag(quay_token, namespace, name, tag["name"])
+        elif tag["name"].endswith(".src"):
+            LOGGER.info("Removing deprecated tag %s", tag["name"])
+            delete_image_tag(quay_token, namespace, name, tag["name"])
```
chmeliik, line 96: Discussed in Slack: we should only remove these if the corresponding binary image doesn't exist

Otherwise, we could delete the only source container tag (i.e. completely delete the source container) for an image (for old images, built before we started pushing the new tag).

tkdchen: Updated. PTAL.

config/registry_image_pruner/image_pruner/test_prune_images.py
```diff
@@ -142,6 +142,23 @@ def test_remove_orphan_tags_with_expected_suffixes(self, urlopen):
...
+                # single deprecated source image tag remains
+                {"name": "123abcd.src", "manifest_digest": "sha256:1234566"},
+
+                # old image has only the deprecated source image, which should not be removed
+                {"name": "donotdelete", "manifest_digest": "sha256:1234567"},
+                {"name": "donotdelete.src", "manifest_digest": "sha256:1234568"},
+
+                # the binary image was deleted before. Now, these should be deleted.
+                {"name": "build-100.src", "manifest_digest": "sha256:1345678"},
+                {"name": "sha256-4567890.src", "manifest_digest": "sha256:1345678"},
```
chmeliik, line 155: I don't quite understand this test case. Was the source tag supposed to be `sha256-1345678.src`?

chmeliik (approved):
LGTM, minor question about tests

config/registry_image_pruner/image_pruner/test_prune_images.py
tkdchen: This aims to be a status that is the binary image `xxx:build-100` has been removed and the pruner starts to clean up.

chmeliik: Ah right, I missed the **.src** in `build-100.src` :+1:

PR #115, opened 2024-05-22
Temporarily disable image pruner

chmeliik:
[STONEBLD-2466](https://issues.redhat.com//browse/STONEBLD-2466)

The pruner may improperly delete the 'sha256-${digest}.*' attachments for images that still exists.

The pruner is not supposed to delete these attachments if the 'sha256:${digest}' image exists. However, it only takes into account tagged images. Platform-specific manifests do not have to be tagged, it is enough if they are part of a multi-platform image index.

Disable the pruner until the issue is fixed properly.

### DIFF
--- a/config/registry_image_pruner/cronjob.yaml
+++ b/config/registry_image_pruner/cronjob.yaml
@@ -27,7 +27,7 @@ spec:
             command:
               - /bin/bash
               - '-c'
-              - python /image-pruner/prune_images.py --namespace $(NAMESPACE)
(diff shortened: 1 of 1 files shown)


rcerven (approved):

chmeliik:
kube-linter is complaining aobut

```
container "manager" does not expose port 8081 for the HTTPGet
object: image-controller-system/image-controller-controller-manager apps/v1, Kind=Deployment
```

I am going to ignore it

PR #116, opened 2024-05-23
Do not delete sig/att/sbom if subject image exists

rcerven:
when getting all tags in registry we are listing only active tags, if tag was removed for the image, image still will be in registry if it is part of image index

if manifest digest isn't present in any of active tags, try query also manifest digest to verify its presence

[STONEBLD-2466](https://issues.redhat.com//browse/STONEBLD-2466)

### DIFF
--- a/config/registry_image_pruner/image_pruner/prune_images.py
+++ b/config/registry_image_pruner/image_pruner/prune_images.py
@@ -79,19 +79,49 @@ def delete_image_tag(quay_token: str, namespace: str, name: str, tag: str) -> No
             raise(ex)
 
 
+def manifest_exists(quay_token: str, namespace: str, name: str, manifest: str) -> bool:
+    api_url = f"{QUAY_API_URL}/repository/{namespace}/{name}/manifest/{manifest}"
+    request = Request(api_url, headers={
+        "Authorization": f"Bearer {quay_token}",
+    })
+    resp: HTTPResponse
+    manifest_exists = True
+    try:
+        with urlopen(request) as resp:
+            if resp.status != 200 and resp.status != 204:
+                raise RuntimeError(resp.reason)
+
+    except HTTPError as ex:
+        if ex.status != 404:
+            raise(ex)
+        else:
+            manifest_exists = False
+
+    return manifest_exists
+
+
 def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, name: str, dry_run: bool = False) -> None:
     image_digests = [image["manifest_digest"] for image in tags]
     tags_map = {tag_info["name"]: tag_info for tag_info in tags}
     tag_regex = re.compile(r"^sha256-([0-9a-f]+)(\.sbom|\.att|\.src|\.sig)$")
+    manifests_checked = {}
     for tag in tags:
         # attestation or sbom image
         if (match := tag_regex.match(tag["name"])) is not None:
             if f"sha256:{match.group(1)}" not in image_digests:
-                if dry_run:
-                    LOGGER.info("Tag %s from %s/%s should be removed", tag["name"], namespace, name)
-                else:
-                    LOGGER.info("Removing tag %s from %s/%s", tag["name"], namespace, name)
-                    delete_image_tag(quay_token, namespace, name, tag["name"])
+                # verify that manifest really doesn't exist, because if tag was removed, it won't be in tag list, but may still be in the registry
+                manifest_existence = manifests_checked.get(f"sha256:{match.group(1)}")
+                if manifest_existence is None:
+                    manifest_existence = manifest_exists(quay_token, namespace, name, f"sha256:{match.group(1)}")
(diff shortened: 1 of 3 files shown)


arewm (approved):

lcarva (approved):
:taco:

config/registry_image_pruner/image_pruner/test_prune_images.py
```diff
@@ -364,7 +394,34 @@ def test_remove_tags_nothing_to_remove(self, delete_image_tag):
...
+        ]
+
+        manifest_exists.side_effect = [
+            True,
+            True,
+        ]
+
+        with self.assertRaisesRegex(AssertionError, expected_regex="no logs of level INFO"):
+            with self.assertLogs(LOGGER) as logs:
+                remove_tags(tags, QUAY_TOKEN, "some", "repository", dry_run=True)
```
tkdchen, line 417: It does not make sense to set `dry_run=True` here. When it is True, `delete_image_tag` will be never called under whatever conditions.

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -87,11 +108,14 @@ def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, nam
         # attestation or sbom image
         if (match := tag_regex.match(tag["name"])) is not None:
             if f"sha256:{match.group(1)}" not in image_digests:
```
tkdchen, line 110: We don't need to make this check. `sha256-*.(sig|att|sbom|src)` tags should not be deleted as long as the image with digest `match.group(1)` exists in the registry.

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -87,11 +108,14 @@ def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, nam
...
-                else:
-                    LOGGER.info("Removing tag %s from %s/%s", tag["name"], namespace, name)
-                    delete_image_tag(quay_token, namespace, name, tag["name"])
+                # verify that manifest really doesn't exist, because if tag was removed, it won't be tag list, but still be in the registry
+                if not manifest_exists(quay_token, namespace, name, f"sha256:{match.group(1)}"):
+                    if dry_run:
+                        LOGGER.info("Tag %s from %s/%s should be removed", tag["name"], namespace, name)
+                    else:
+                        LOGGER.info("Removing tag %s from %s/%s", tag["name"], namespace, name)
+                        delete_image_tag(quay_token, namespace, name, tag["name"])
```
tkdchen, line 117: The existent or nonexistent manifest digests can be cached to reduce the number of HTTP requests to Quay.io.

chmeliik: +1, currently we're querying the same manifest four times - once for each `sha256-${digest}.(att|sig|sbom|src)`

config/registry_image_pruner/image_pruner/prune_images.py
chmeliik: On the other hand, it could be a nice optimization. If the digest _is_ in the tag list, we don't have to query the API at all.

config/registry_image_pruner/image_pruner/prune_images.py
```diff
@@ -87,11 +108,14 @@ def remove_tags(tags: List[Dict[str, Any]], quay_token: str, namespace: str, nam
         # attestation or sbom image
         if (match := tag_regex.match(tag["name"])) is not None:
             if f"sha256:{match.group(1)}" not in image_digests:
-                if dry_run:
-                    LOGGER.info("Tag %s from %s/%s should be removed", tag["name"], namespace, name)
-                else:
-                    LOGGER.info("Removing tag %s from %s/%s", tag["name"], namespace, name)
-                    delete_image_tag(quay_token, namespace, name, tag["name"])
+                # verify that manifest really doesn't exist, because if tag was removed, it won't be tag list, but still be in the registry
```
chmeliik, line 111:
```suggestion
                # verify that manifest really doesn't exist, because if tag was removed, it won't be in tag list, but may still be in the registry
```

config/registry_image_pruner/image_pruner/test_prune_images.py
rcerven: well it was in that other test above :)

config/registry_image_pruner/image_pruner/prune_images.py
rcerven: well it is tag list, and that tag is in tag list when not using 'activeTagsOnly', but that also shows all removed tags ....

rcerven: oh right, yes that should handle most cases

rcerven:
added caching, so we will query manifest only once

chmeliik (approved):

PR #119, opened 2024-06-04
Fix migration script to attach push secrets to SA

mshaposhnik:
Fix push/pull secrets migration script to attach push secrets to `appstudio-pipeline` SA

### DIFF
--- a/hack/remove_rs.sh
+++ b/hack/remove_rs.sh
@@ -33,6 +33,8 @@ for QN in "${LIST[@]}"; do
       exit 1
     fi
 
+    sleep 3
+
      # re-create secret from json
    if echo $SJSON | kubectl create -f -
     then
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

hack/remove_rs.sh
```diff
@@ -40,6 +42,17 @@ for QN in "${LIST[@]}"; do
       exit 1
     fi
 
+    #patch service account for push secrets
+    if [[ $NAME == *-push ]]; then
+      if kubectl patch serviceaccount $SA_NAME -n $NS -p '{"secrets": [{"name": "'$NAME'"}]}' --type="merge"
```
mmorhun, line 47: I think we need it in pull-secrets section as well

mshaposhnik: If I understood correct, must be like that:
```
secrets:
  - name: devfile-sample-go-basic-push
  - name: appstudio-pipeline-dockercfg-n6rvr
imagePullSecrets:
  - name: devfile-sample-go-basic-pull
  - name: appstudio-pipeline-dockercfg-n6rvr

```

mmorhun (approved):

PR #122, opened 2024-06-07
Quay notification add, update and delete support for ImageRepository

mavaras:
### Description
Extends the controller to be able to react to CR updates and handle the notifications accordingly.
It allows to modify existing CR specification by adding new notifications, modifying existing ones and deleting them. The controller will react to those changes and reflect them accordingly in Quay as well as in the CR status.

### How has this been tested?
Quay's _UpdateNotification_ and _DeleteNotification_ tested on a mock personal Quay repository using the _quay_test.go_ suite.
Whole functionality tested by deploying the operator to an Openshift cluster and using a test Quay organization to play with ImageRepository's notifications by testing all the functional workflows.

---
This PR is a continuation of this one: https://github.com/konflux-ci/image-controller/pull/112
JIRA: [ISV-4914](https://issues.redhat.com/browse/ISV-4914)

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -218,6 +218,15 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 	}
 
+	if err = r.HandleNotifications(ctx, imageRepository); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.Client.Status().Update(ctx, imageRepository); err != nil {
+		log.Error(err, "failed to update image repository status")
+		return ctrl.Result{}, err
+	}
+
 	// we are adding to map only for new provision, not for some partial actions,
 	// so report time only if time was recorded
 	provisionTime, timeRecorded := metrics.RepositoryTimesForMetrics[repositoryIdForMetrics]
@@ -228,51 +237,6 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	delete(metrics.RepositoryTimesForMetrics, repositoryIdForMetrics)
 
 	return ctrl.Result{}, nil
-}
-
-func (r *ImageRepositoryReconciler) AddNotifications(ctx context.Context, imageRepository *imagerepositoryv1alpha1.ImageRepository) ([]imagerepositoryv1alpha1.NotificationStatus, error) {
-	log := ctrllog.FromContext(ctx).WithName("ConfigureNotifications")
-
-	if imageRepository.Spec.Notifications == nil {
-		// No notifications to configure
-		return nil, nil
-	}
-
-	log.Info("Configuring notifications")
-	notificationStatus := []imagerepositoryv1alpha1.NotificationStatus{}
-
-	for _, notification := range imageRepository.Spec.Notifications {
-		log.Info("Creating notification in Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
-		quayNotification, err := r.QuayClient.CreateNotification(
-			r.QuayOrganization,
-			imageRepository.Spec.Image.Name,
-			quay.Notification{
-				Title:  notification.Title,
-				Event:  string(notification.Event),
-				Method: string(notification.Method),
-				Config: quay.NotificationConfig{
-					Url: notification.Config.Url,
-				},
-				EventConfig: quay.NotificationEventConfig{},
-			})
-		if err != nil {
-			log.Error(err, "failed to create notification", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
-			return nil, err
-		}
-		notificationStatus = append(
-			notificationStatus,
-			imagerepositoryv1alpha1.NotificationStatus{
-				UUID:  quayNotification.UUID,
-				Title: notification.Title,
-			})
-
-		log.Info("Notification added",
-			"Title", notification.Title,
-			"Event", notification.Event,
-			"Method", notification.Method,
-			"QuayNotification", quayNotification)
-	}
-	return notificationStatus, nil
 }
 
 // ProvisionImageRepository creates image repository, robot account(s) and secret(s) to access the image repository.
@@ -364,7 +328,7 @@ func (r *ImageRepositoryReconciler) ProvisionImageRepository(ctx context.Context
 	}
 
 	var notificationStatus []imagerepositoryv1alpha1.NotificationStatus
-	if notificationStatus, err = r.AddNotifications(ctx, imageRepository); err != nil {
+	if notificationStatus, err = r.SetNotifications(ctx, imageRepository); err != nil {
 		return err
 	}
 
@@ -683,4 +647,15 @@ func getRandomString(length int) string {
 		panic("Failed to read from random generator")
 	}
 	return hex.EncodeToString(bytes)[0:length]
+}
+
+func (r *ImageRepositoryReconciler) UpdateImageRepositoryStatusMessage(ctx context.Context, imageRepository *imagerepositoryv1alpha1.ImageRepository, statusMessage string) error {
+	log := ctrllog.FromContext(ctx)
+	imageRepository.Status.Message = statusMessage
+	if err := r.Client.Status().Update(ctx, imageRepository); err != nil {
+		log.Error(err, "failed to update image repository status")
+		return err
+	}
+
+	return nil
 }
--- a/controllers/imagerepository_controller_test.go
+++ b/controllers/imagerepository_controller_test.go
@@ -19,13 +19,13 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/konflux-ci/image-controller/pkg/quay"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	"regexp"
 	"strings"
 	"time"
 
-	"github.com/konflux-ci/image-controller/pkg/quay"
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -315,6 +315,21 @@ var _ = Describe("Image repository controller", func() {
 				Expect(organization).To(Equal(quay.TestQuayOrg))
 				return &quay.Notification{UUID: "uuid"}, nil
 			}
+			isGetNotificationsInvoked := false
+			quay.GetNotificationsFunc = func(organization, repository string) ([]quay.Notification, error) {
+				isGetNotificationsInvoked = true
+				Expect(organization).To(Equal(quay.TestQuayOrg))
+				return []quay.Notification{
+					{
+						Title:  "test-notification",
+						Event:  string(imagerepositoryv1alpha1.NotificationEventRepoPush),
+						Method: string(imagerepositoryv1alpha1.NotificationMethodWebhook),
+						Config: quay.NotificationConfig{
+							Url: "http://test-url",
+						},
+					},
+				}, nil
+			}
 
 			imageRepositoryConfigObject := imageRepositoryConfig{
 				Labels: map[string]string{
@@ -354,6 +369,7 @@ var _ = Describe("Image repository controller", func() {
 			if updateComponentAnnotation {
 				Expect(component.Spec.ContainerImage).To(Equal(imageRepository.Status.Image.URL))
 				Expect(imageRepository.Annotations).To(HaveLen(0))
+				Eventually(func() bool { return isGetNotificationsInvoked }, timeout, interval).Should(BeTrue())
 			} else {
 				Expect(component.Spec.ContainerImage).To(BeEmpty())
 			}
@@ -455,6 +471,21 @@ var _ = Describe("Image repository controller", func() {
 				isRegenerateRobotAccountTokenForPushInvoked = true
 				return &quay.RobotAccount{Name: robotName, Token: newPushToken}, nil
 			}
+			isGetNotificationsInvoked := false
+			quay.GetNotificationsFunc = func(organization, repository string) ([]quay.Notification, error) {
+				isGetNotificationsInvoked = true
+				Expect(organization).To(Equal(quay.TestQuayOrg))
+				return []quay.Notification{
+					{
+						Title:  "test-notification",
+						Event:  string(imagerepositoryv1alpha1.NotificationEventRepoPush),
+						Method: string(imagerepositoryv1alpha1.NotificationMethodWebhook),
+						Config: quay.NotificationConfig{
+							Url: "http://test-url",
+						},
+					},
+				}, nil
+			}
 
 			imageRepository := getImageRepository(resourceKey)
 			oldTokenGenerationTimestamp := *imageRepository.Status.Credentials.GenerationTimestamp
@@ -470,6 +501,7 @@ var _ = Describe("Image repository controller", func() {
 					imageRepository.Status.Credentials.GenerationTimestamp != nil &&
 					*imageRepository.Status.Credentials.GenerationTimestamp != oldTokenGenerationTimestamp
 			}, timeout, interval).Should(BeTrue())
+			Eventually(func() bool { return isGetNotificationsInvoked }, timeout, interval).Should(BeTrue())
 
 			pushSecretKey := types.NamespacedName{Name: imageRepository.Status.Credentials.PushSecretName, Namespace: imageRepository.Namespace}
 			pushSecret := waitSecretExist(pushSecretKey)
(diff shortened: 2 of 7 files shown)


controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,71 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		// No status notifications to check
+		return nil
+	}
+
+	log.Info("Handling notifications")
+	allNotifications, err := r.QuayClient.GetNotifications(r.QuayOrganization, imageRepository.Spec.Image.Name)
+	if err != nil {
+		return err
+	}
+	for index, statusNotification := range imageRepository.Status.Notifications {
```
Allda, line 250: This section seems to deserve some refactoring to avoid so many nested layers of `for` loops and `if` statements. :D

mavaras: I couldn't find any other way to do all the comparisons between Spec.Notifications, Status.Notifications and Quay notifications :thinking:

controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,139 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
+func (r *ImageRepositoryReconciler) NotificationExistsInQuayByUUID(UUID string, imageRepository *imagerepositoryv1alpha1.ImageRepository) (quay.Notification, error) {
```
mavaras, line 237: Probably it exists a more elegant way to have this two functions (or not). Let me know any idea.

controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,139 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+							Event:  string(notification.Event),
+							Method: string(notification.Method),
+							Config: quay.NotificationConfig{
+								Url: notification.Config.Url,
+							},
+							EventConfig: quay.NotificationEventConfig{},
+						})
+					if err != nil {
+						log.Error(err, "failed to update notification", "Title", statusNotification.Title, "UUID", statusNotification.UUID)
+						return err
```
mmorhun, line 305: It will end up in an infinite reconcile loop, since spec is not changed.

controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,139 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			if notification.Title == statusNotification.Title {
+				existsInSpec = true
+				quayNotification, err := r.NotificationExistsInQuayByUUID(statusNotification.UUID, imageRepository)
+				if err != nil {
+					return err
+				}
+				if quayNotification.UUID != "" && (quayNotification.Title != notification.Title || quayNotification.Event != string(notification.Event) || quayNotification.Method != string(notification.Method) || quayNotification.Config.Url != notification.Config.Url) {
+					// Updated item in Spec.Notifications: update notification in Quay
+					log.Info("Updating notification in Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+					updatedNotification, err := r.QuayClient.UpdateNotification(
```
mmorhun, line 290: This is really similar to what's done in `AddNotification`

controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,139 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+	if err := r.Client.Status().Update(ctx, imageRepository); err != nil {
+		log.Error(err, "failed to update image repository status")
+		return err
+	}
+
+	return nil
+}
+
+func (r *ImageRepositoryReconciler) AddNotification(notification imagerepositoryv1alpha1.Notifications, imageRepository *imagerepositoryv1alpha1.ImageRepository) (imagerepositoryv1alpha1.NotificationStatus, error) {
```
mmorhun, line 350: Should we do any validation before sending a request to Quay?

pkg/quay/quay.go
```diff
@@ -42,6 +42,8 @@ type QuayService interface {
 	DeleteTag(organization, repository, tag string) (bool, error)
 	GetNotifications(organization, repository string) ([]Notification, error)
 	CreateNotification(organization, repository string, notification Notification) (*Notification, error)
+	UpdateNotification(organization, repository string, notificationUuid string, notification Notification) (*Notification, error)
```
mmorhun, line 45: Why not to use pointer to `Notification`?

pkg/quay/quay_debug_test.go
```diff
@@ -231,5 +231,101 @@ func TestCreateNotification(t *testing.T) {
 	if !found {
 		t.Fatalf("Notification %s not found", quayNotification.UUID)
 	}
+}
+
+func TestDeleteNotification(t *testing.T) {
```
mmorhun, line 236: It's not a unit test, please create a test per quay client method as it's done in this file.

mmorhun (changes requested):
Please see the remarks to fix.
Also, there are a lot of logic related to notifications, I think we need to have a separate file with all notifications logic in it.

controllers/imagerepository_controller.go
mavaras: Note that here we are **Updating** the notification, which is a different operation to **Add** the notification, although the calls could look similar

controllers/imagerepository_controller.go
mavaras: Let me check that with Ales. Here, 100% I just refactored the code that was already written.

pkg/quay/quay_debug_test.go
mavaras: I don't get your point. I included here _TestDeleteNotification_ and _TestUpdateNotification_ the same as _TestCreateNotification_. They do the same as _TestCreateNotification_ does but for delete and update operations and check several output cases to verify it works well.

controllers/imagerepository_controller.go
```diff
@@ -230,6 +234,139 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		return nil
+	}
+	log.Info("Handling notifications")
+	for index := 0; index < len(imageRepository.Status.Notifications); index++ {
+		statusNotification := &imageRepository.Status.Notifications[index] // This way we can update the UUID if notification gets updated
+		existsInSpec := false
+		for _, notification := range imageRepository.Spec.Notifications {
+			if notification.Title == statusNotification.Title {
+				existsInSpec = true
+				quayNotification, err := r.NotificationExistsInQuayByUUID(statusNotification.UUID, imageRepository)
```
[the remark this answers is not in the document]
mavaras: I get your point. Let me think on an alternative here.

mavaras:
> Please see the remarks to fix. Also, there are a lot of logic related to notifications, I think we need to have a separate file with all notifications logic in it.

Thanks for your comments Mykola. I agree on the notifications module to handle all related to it separately. I can take care of that in a next PR, as it is not part of the ACs that I am working on. We can discuss how to create it these days :+1:

controllers/imagerepository_controller.go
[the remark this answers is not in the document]
mavaras: Note that call to _NotificationExistsInQuayByUUID_ or _NotificationExistsInQuayByTitle_ is performed only once per iteration (= per notification). It depends on the if else where they are placed. It couldn't happen that we call both in the same iteration.

Besides that, you are 100% right in the fact that we should only call Quay's _getNotifications_ only once per _HandleNotifications_ function call. I'll refactor it to avoid call it several times and let _NotificationExistsInQuayByUUID_ and _NotificationExistsInQuayByTitle_ only do comparisons based on the attribute and not calling _getNotifications_ within them.

mmorhun:
> I can take care of that in a next PR

@mavaras it just moving it into a new file `imagerepository_notifications.go`, very simple, no need in a separate go module.

mavaras:
PR updated with the following changes:
- Notifications file created containing all notifications-related functions.
- Avoid to block the `Reconcile` loop by returning errors. `UpdateImageRepositoryStatus` function created to simplify the CR status update.
- Reduce logs cadence

All the changes has been tested again in the same way as described in the "_How has this been tested?_" section of this PR description.

@mmorhun PTAL
Thanks

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Info("Notification added",
+			"Title", notification.Title,
+			"Event", notification.Event,
+			"Method", notification.Method,
+			"QuayNotification", notificationStatusRes)
+	}
+	return notificationStatus, nil
+}
+
+func NotificationExistsInQuayByUUID(UUID string, imageRepository *imagerepositoryv1alpha1.ImageRepository, allNotifications []quay.Notification) (quay.Notification, error) {
```
mmorhun, line 160: I wouldn't export these functions (start the name from lower case letter)

mmorhun (changes requested):

pkg/quay/quay_debug_test.go
mmorhun: They are not tests, it's to check if the client implementation works as expected. You can have a smoke test function to test all the notifications cases, but this file is designed to invoke only one method per run (mainly for debug purposes), for example, see how robot accounts handled.
If there is a problem, say, with delete notification, you won't be able to debug it without changing the function (otherwise it will fail on create notification above or it would create another notification that's anyway no desired).

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	for index := 0; index < len(imageRepository.Status.Notifications); index++ {
+		statusNotification := &imageRepository.Status.Notifications[index] // This way we can update the UUID if notification gets updated
+		existsInSpec := false
+		for _, notification := range imageRepository.Spec.Notifications {
+			if notification.Title == statusNotification.Title {
+				existsInSpec = true
+				quayNotification, err := NotificationExistsInQuayByUUID(statusNotification.UUID, imageRepository, allNotifications)
+				if err != nil {
+					if err = r.UpdateImageRepositoryStatus(ctx, imageRepository, "Notification does not exists in Quay (by UUID)"); err != nil {
```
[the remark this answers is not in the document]
mavaras: 1. You are right. The status message is definitely not accurate. I'll change it to something more realistic.
2. In that case the condition in L49 (`if notification.Title == statusNotification.Title {`) wouldn't be true (no notification in image repository status) and we get to L83 where we add the notification to Quay.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+		// No notifications to configure
+		return nil, nil
+	}
+
+	log.Info("Configuring notifications")
+	notificationStatus := []imagerepositoryv1alpha1.NotificationStatus{}
+
+	for _, notification := range imageRepository.Spec.Notifications {
+		log.Info("Creating notification in Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+		notificationStatusRes, err := r.AddNotification(notification, imageRepository)
```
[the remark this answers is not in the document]
mavaras: If the notification already exists in Quay then Quay API wouldn't return an error (err == nil).
I think this in intended. As I mentioned in another previous comment in this PR, this code is just refactored by me, you can check that this is done exactly as this here: https://github.com/konflux-ci/image-controller/blob/main/controllers/imagerepository_controller.go#L258
And also, that code (AddNotifications etc.) was reviewed and approved in the previous PR to this one (also mentioned): https://github.com/konflux-ci/image-controller/pull/112

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		if err = r.UpdateImageRepositoryStatus(ctx, imageRepository, "Couldn't retrieve all Quay notifications"); err != nil {
+			return err
+		}
+		return nil
+	}
+	for index := 0; index < len(imageRepository.Status.Notifications); index++ {
+		statusNotification := &imageRepository.Status.Notifications[index] // This way we can update the UUID if notification gets updated
+		existsInSpec := false
+		for _, notification := range imageRepository.Spec.Notifications {
```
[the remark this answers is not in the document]
mavaras: You can see here how the structure of Spec.Notifications and Status.Notifications is: https://github.com/konflux-ci/image-controller/pull/112#issue-2302454077

Only Status.Notifications have UUID, the piece required to interact with Quay API to do operations such as **Update** and **Delete**. At the end, we have to compare all Status.Notifications items with the ones in Spec.Notifications and when we do not have a match, we check if that's because a new item is present in Spec.Notifications and in that case we do the **Add** operation.

Since the UUID is present in Status.Notifications it makes more sense to me (and is more legible/structured) to compare that items with the ones in Spec.Notifications (+ we could have less loop iterations by starting with Status.Notifications items)

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+		if quayNotification.UUID == UUID {
+			notification = quayNotification
+			break
+		}
+	}
+
+	return notification, nil
+}
+
+func NotificationExistsInQuayByTitle(Title string, imageRepository *imagerepositoryv1alpha1.ImageRepository, allNotifications []quay.Notification) (quay.Notification, error) {
```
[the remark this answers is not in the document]
mavaras: I found a way of refactoring the calls to keep the function as it is. Thanks for pointing this out.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,186 @@
...
+
+import (
+	"context"
+
+	imagerepositoryv1alpha1 "github.com/konflux-ci/image-controller/api/v1alpha1"
+	"github.com/konflux-ci/image-controller/pkg/quay"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func (r *ImageRepositoryReconciler) AddNotification(notification imagerepositoryv1alpha1.Notifications, imageRepository *imagerepositoryv1alpha1.ImageRepository) (imagerepositoryv1alpha1.NotificationStatus, error) {
```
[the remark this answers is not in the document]
mmorhun: It's still not fixed. If a user input `blah` as method or something, Quay will probably return an error. Then, the error will be returned [here](https://github.com/konflux-ci/image-controller/blob/main/controllers/imagerepository_controller.go#L370) and up to the reconcile. Then retry, endless loop.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+			EventConfig: quay.NotificationEventConfig{},
+		})
+
+	if err != nil {
+		return imagerepositoryv1alpha1.NotificationStatus{}, err
+	}
+	return imagerepositoryv1alpha1.NotificationStatus{
+		UUID:  quayNotification.UUID,
+		Title: notification.Title,
+	}, err
```
mmorhun, line 31: nitpick: since you handle error above, you can have `nil` here.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+						if err = r.UpdateImageRepositoryStatusMessage(ctx, imageRepository, "Error while updating the notification in Quay"); err != nil {
+							return err
+						}
+						return nil
+					}
+					statusNotification.UUID = updatedNotification.UUID
+				}
+			} else {
+				quayNotification := notificationExistsInQuayByTitle(notification.Title, imageRepository, allNotifications)
+				if quayNotification.UUID == "" {
```
mmorhun, line 82: If the notification does not exists, it's better to return `nil` from `notificationExistsInQuayByTitle`. Then `quayNotification.UUID == ""` will not be ambiguous.
Moreover, if found notification data is not used, it's even better to return `bool`

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+					// New item in Spec.Notifications: add notification to Quay and Status.Notifications
+					log.Info("Adding new notification to Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+					notificationStatus, err := r.AddNotification(notification, imageRepository)
+					if err != nil {
+						log.Error(err, "failed to add notification", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+						if err = r.UpdateImageRepositoryStatusMessage(ctx, imageRepository, "Error while adding a notification to Quay"); err != nil {
```
mmorhun, line 88: I think title will be useful in log message.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+							return err
+						}
+						return nil
+					}
+					imageRepository.Status.Notifications = append(imageRepository.Status.Notifications, notificationStatus)
+				}
+			}
+		}
+		if existsInSpec == false {
```
mmorhun, line 97: `if !existsInSpec {`

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+			// Remove notification from CR status
+			imageRepository.Status.Notifications = append(imageRepository.Status.Notifications[:index], imageRepository.Status.Notifications[index+1:]...)
+		}
+	}
+
+	return nil
+}
+
+func (r *ImageRepositoryReconciler) AddNotifications(ctx context.Context, imageRepository *imagerepositoryv1alpha1.ImageRepository) ([]imagerepositoryv1alpha1.NotificationStatus, error) {
```
mmorhun, line 119: It overwrites all existing notifications in status. So, the name is confusing. Please rename or/and add a comment that explains the behavior of this method.

pkg/quay/quay_debug_test.go
```diff
@@ -231,5 +231,101 @@ func TestCreateNotification(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if quayNotification == nil {
+		t.Fatal("Notification should not be nil")
+	}
+	if quayNotification.UUID == "" {
+		t.Fatal("Notification UUID should not be empty")
+	}
+
+	res, err := quayClient.DeleteNotification(quayOrgName, quayImageRepoName, quayNotification.UUID)
```
mmorhun, line 262: This should be in another `Test****`

pkg/quay/quay_debug_test.go
```diff
@@ -231,5 +231,101 @@ func TestCreateNotification(t *testing.T) {
...
+	if err != nil || res == false {
+		t.Fatal(err)
+	}
+
+	allNotifications, err := quayClient.GetNotifications(quayOrgName, quayImageRepoName)
```
mmorhun, line 267: This should be in another `Test****`

pkg/quay/quay_debug_test.go
```diff
@@ -231,5 +231,101 @@ func TestCreateNotification(t *testing.T) {
...
+			found = true
+			break
+		}
+	}
+	if found {
+		t.Fatalf("Notification %s found after deletion", quayNotification.UUID)
+	}
+}
+
+func TestUpdateNotification(t *testing.T) {
```
mmorhun, line 284: Update, in the implementation, is delete and create. We have cases for both above. I see no sense in this function.

controllers/imagerepository_notifications.go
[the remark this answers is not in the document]
mmorhun: Agree, thank you for the explanation.

controllers/imagerepository_notifications.go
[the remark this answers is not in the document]
mavaras: If Quay API receives a POST request with a notification with a wrong field it returns a 400 error with a descriptive error message. I guess this is considered an "intermittent error" as you named it and f.e. Quay API being down is a "persistent error". That's okay.

I understand your point of the reconcile loop but it is not clear what you want here. Please explain **exactly** how you mean to handle those errors.
You mean f.e. to "ignore" the 400 error from Quay API and **not** return an error so the reconcile loop keeps moving? From what you say, the goal would be to not return an error that could propagate to the reconcile loop. With "proper validation of the user input" you mean that we do perform the validation of the notification structure? Why doing that twice? That's exactly what Quay API does and accordingly returns an error + message.

Plus, I'll add again that this piece of code is a refactor from code that already [exists](https://github.com/konflux-ci/image-controller/blob/main/controllers/imagerepository_controller.go#L260), is currently working as this in image-controller codebase and that was [approved](https://github.com/konflux-ci/image-controller/pull/112).

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+					log.Info("Adding new notification to Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+					notificationStatus, err := r.AddNotification(notification, imageRepository)
+					if err != nil {
+						log.Error(err, "failed to add notification", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+						if err = r.UpdateImageRepositoryStatusMessage(ctx, imageRepository, "Error while adding a notification to Quay"); err != nil {
+							return err
+						}
+						return nil
+					}
+					imageRepository.Status.Notifications = append(imageRepository.Status.Notifications, notificationStatus)
```
Allda, line 93: You update the notification array while iterating over it in the `for` loop. I am not familiar with how Golang handles this situation but other languages usually end up in an unpredictable state. Is this operation safe in golang?

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,173 @@
...
+				statusNotification.UUID)
+			if err != nil {
+				log.Error(err, "failed to delete notification", "Title", statusNotification.Title, "UUID", statusNotification.UUID)
+				if err = r.UpdateImageRepositoryStatusMessage(ctx, imageRepository, "Error while deleting the notification from Quay"); err != nil {
+					return err
+				}
+				return nil
+			}
+			// Remove notification from CR status
+			imageRepository.Status.Notifications = append(imageRepository.Status.Notifications[:index], imageRepository.Status.Notifications[index+1:]...)
```
Allda, line 112: Same here. You update the array inside the for loop.

controllers/imagerepository_notifications.go
mavaras: Yes, I took that into consideration. I verified it woks by using a system like the one in L49 to update some fields and to reflect the status updates it works by updating the ImageRepository status with `r.Client.Status().Update(ctx, imageRepository)` calls. I noticed that same strategy is followed in different parts of the codebase.

mavaras: At the beginning the status update call was placed inside `HandleNotifications` but after Mykola's suggestions I moved it right after calling the function in the reconcile loop.

Allda:
I have tested the new controller and it seems it supports all use cases and all notification modifications in CR are propagated to the Quay.

controllers/imagerepository_notifications.go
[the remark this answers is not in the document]
mmorhun: My main point is to avoid endless error loop. Returning or not an error is an implementation detail.
To properly solve the problem we need to distinguish between intermittent (a retry could help) and persistent (retry won't help) errors.

> If Quay API receives a POST request with a notification with a wrong field it returns a 400 error with a descriptive error message. I guess this is considered an "intermittent error" 

It's persistent error, since retry will not change the result/outcome (unlike in case of a network error).

> Plus, I'll add again that this piece of code is a refactor from code that already [exists](https://github.com/konflux-ci/image-controller/blob/main/controllers/imagerepository_controller.go#L260)

We've merged that code because we had tight deadline. It was promised to make it complete and clean.

controllers/imagerepository_notifications.go
mmorhun: The error still doesn't show which notification creation failed. If it's due to typo in config, as a user, I'd like to know which notification to check (instead of looking through all of them).

controllers/imagerepository_notifications.go
mmorhun: Maybe we should raname it to `SetNotifications`?

mmorhun:
@mavaras also we are missing integration tests in `imagerepository_controller_test.go` for adding/deleting and editing notifications after creation of the ImagreRepository CR.

mavaras:
PR updated with the following changes:

- Integration tests for add/update/delete notifications added in `imagerepository_controller_test.go`
- Improved error handling for Quay-related functions
- Other minor refactors that were required

All the changes have been tested again in the same way as described in the "How has this been tested?" section of this PR description.

controllers/imagerepository_notifications.go
rcerven: this is skipping 1 element of imageRepository.Status.Notifications, because you remove one element, the rest is moved and now that 1st element which should be processed is considered processed 

verified with simple code:

	ar = []int{1, 2, 3, 4, 5, 6}
	fmt.Println(ar)
	for index := 0; index < len(ar); index++ {
		element := &ar[index]
		fmt.Println(*element)
		if index == 1 {
			ar = append(ar[:index], ar[index+1:]...)
		}
	}


output is :

[1 2 3 4 5 6]
1
2
4
5
6

so '3' is skipped because it has now (after append) index which was already processed

mavaras: You're right. I didn't modify the index at the same time (`index--`). Thanks for pointing this out.

pkg/quay/quay_test.go
```diff
@@ -1470,6 +1470,161 @@ func TestQuayClient_CreateNotification(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			defer gock.Off()
+
+			req := gock.New(testQuayApiUrl).
+				MatchHeader("Content-Type", "application/json").
+				MatchHeader("Authorization", "Bearer authtoken").
+				Delete(fmt.Sprintf("repository/%s/%s/notification/%s", org, repo, tc.notificationUuid))
+			req.Reply(tc.statusCode).JSON(tc.response)
+
+			if tc.name == "stop if http request fails" {
```
rcerven, line 1513: this isn't any test case can be removed

pkg/quay/quay_test.go
```diff
@@ -1470,6 +1470,161 @@ func TestQuayClient_CreateNotification(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			defer gock.Off()
+
+			reqDelete := gock.New(testQuayApiUrl).
+				MatchHeader("Content-Type", "application/json").
+				MatchHeader("Authorization", "Bearer authtoken").
+				Delete(fmt.Sprintf("repository/%s/%s/notification/%s", org, repo, tc.notificationUuid))
+			reqDelete.Reply(tc.statusCodeDelete).JSON(tc.responseDelete)
+
+			if tc.name == "stop if http request fails" {
```
rcerven, line 1583: this isn't any test case can be removed

pkg/quay/quay_debug_test.go
```diff
@@ -212,24 +214,37 @@ func TestCreateNotification(t *testing.T) {
 	if quayNotification.UUID == "" {
 		t.Fatal("Notification UUID should not be empty")
 	}
+}
 
-	allNotifications, err := quayClient.GetNotifications(quayOrgName, quayImageRepoName)
```
rcerven, line 216: trying to GetNotifications should be after each method

create/update expecting some to exist
delete  expecting non to exist

mavaras: That was originally present in the code but removed due to a code review. I'll add it back

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,189 @@
...
+				}
+				if !exists {
+					// New item in Spec.Notifications: add notification to Quay and Status.Notifications
+					log.Info("Adding new notification to Quay", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+					notificationStatus, err := r.AddNotification(notification, imageRepository)
+					if err != nil {
+						log.Error(err, "failed to add notification", "Title", notification.Title, "Event", notification.Event, "Method", notification.Method)
+						return r.handleError(ctx, imageRepository, err, "Error while adding a notification ("+notification.Title+") to Quay")
+					}
+					imageRepository.Status.Notifications = append(imageRepository.Status.Notifications, notificationStatus)
```
rcerven, line 90: you should also check here IF the notitification isn't already in the status and add it only if it is missing, to avoid duplicity in status

pkg/quay/quay_test.go
```diff
@@ -1470,6 +1470,153 @@ func TestQuayClient_CreateNotification(t *testing.T) {
...
+				Delete(fmt.Sprintf("repository/%s/%s/notification/%s", org, repo, tc.notificationUuid))
+			reqDelete.Reply(tc.statusCodeDelete).JSON(tc.responseDelete)
+
+			reqCreate := gock.New(testQuayApiUrl).
+				MatchHeader("Content-Type", "application/json").
+				MatchHeader("Authorization", "Bearer authtoken").
+				Post(fmt.Sprintf("repository/%s/%s/notification/", org, repo))
+			reqCreate.Reply(tc.statusCodeCreate).JSON(tc.responseCreate)
+
+			if tc.name == "stop if http request fails" {
```
rcerven, line 1585: can be removed

controllers/imagerepository_notifications.go
mavaras: Discussed via Slack. Concluded to update PR for safety reasons.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (r *ImageRepositoryReconciler) handleError(ctx context.Context, imageRepository *imagerepositoryv1alpha1.ImageRepository, err error, errorStatusMessage string) error {
+	if strings.Contains(err.Error(), "400") || strings.Contains(err.Error(), "404") {
+		if err = r.UpdateImageRepositoryStatusMessage(ctx, imageRepository, errorStatusMessage); err != nil {
+			return err
+		}
+		return nil
+	} else {
+		return err
```
mmorhun, line 196: You can just return without else block.

controllers/imagerepository_notifications.go
```diff
@@ -0,0 +1,198 @@
...
+			"Event", notification.Event,
+			"Method", notification.Method,
+			"QuayNotification", notificationStatusRes)
+	}
+	return notificationStatus, nil
+}
+
+func (r *ImageRepositoryReconciler) notificationExistsInQuayByUUID(UUID string, imageRepository *imagerepositoryv1alpha1.ImageRepository) (quay.Notification, error) {
+	notification := quay.Notification{}
+	allNotifications, err := r.QuayClient.GetNotifications(r.QuayOrganization, imageRepository.Spec.Image.Name)
```
mmorhun, line 155: This is called at each iteration, why have you reverted fetching all notifications before the loop?

mavaras: Explained via Slack right now

controllers/imagerepository_controller_test.go
```diff
@@ -531,6 +563,273 @@ var _ = Describe("Image repository controller", func() {
...
+
+		It("should add notification", func() {
+			isCreateNotificationInvoked := false
+			quay.CreateNotificationFunc = func(organization, repository string, notification quay.Notification) (*quay.Notification, error) {
+				isCreateNotificationInvoked = true
+				Expect(organization).To(Equal(quay.TestQuayOrg))
+				return &quay.Notification{UUID: "uuid2", Title: "test-notification-2"}, nil
+			}
+			isGetNotificationsInvoked := false
+			count := 0
```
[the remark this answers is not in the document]
mavaras: Explained via Slack right now

controllers/imagerepository_notifications.go
mmorhun: Discussed in a meeting, count removed.

mmorhun: Discussed in a meeting, the approach has to be optimized.

controllers/imagerepository_controller_test.go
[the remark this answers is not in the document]
mmorhun: Discussed in a meeting, count variable is deleted.

mmorhun (approved):

mmorhun:
@mavaras thank you for the contribution!

PR #126, opened 2024-06-24
old component_image_reconciler now only creates ImageRepository CR

rcerven:
for backwards compatibility

[STONEBLD-2460](https://issues.redhat.com//browse/STONEBLD-2460)


controllers/component_image_controller.go
```diff
@@ -113,53 +107,17 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-			}
-
-			if err := r.Client.Get(ctx, req.NamespacedName, component); err != nil {
-				log.Error(err, "failed to get Component", l.Action, l.ActionView)
-				return ctrl.Result{}, err
-			}
 			controllerutil.RemoveFinalizer(component, ImageRepositoryComponentFinalizer)
 			if err := r.Client.Update(ctx, component); err != nil {
-				log.Error(err, "failed to remove image repository finalizer", l.Action, l.ActionUpdate)
+				log.Error(err, "failed to remove image repository finalizer", l.Action, l.ActionUpdate, "componentName", component.Name)
```
mmorhun, line 112: Rather than checking if all components are cleaned and then clean this code, maybe it's easier to run a migration script that just removes this finalizers? But up to you.

controllers/component_image_controller.go
```diff
@@ -187,119 +145,76 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, r.reportError(ctx, component, message)
 	}
 
-	setMetricsTime(componentIdForMetrics, reconcileStartTime)
```
mmorhun, line 190: Should we remove this metric at all?

controllers/component_image_controller.go
```diff
@@ -187,119 +145,76 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	// Update component with the generated data and add finalizer
-	err = r.Client.Get(ctx, req.NamespacedName, component)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("error reading component: %w", err)
-	}
-	if component.ObjectMeta.DeletionTimestamp.IsZero() {
-		component.Annotations[ImageAnnotationName] = string(repositoryInfoBytes)
-		delete(component.Annotations, GenerateImageAnnotationName)
+	if imageRepositoryFound != "" {
+		log.Info("Image repository already exists", "ImageRepositoryName", imageRepositoryFound)
```
mmorhun, line 166: Do we log related component anywhere?

controllers/component_image_controller.go
```diff
@@ -187,119 +145,76 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+	log.Info("Image repository created", "ImageRepositoryName", imageRepositoryName)
+
+	delete(component.Annotations, GenerateImageAnnotationName)
+
+	if err := r.Client.Update(ctx, component); err != nil {
+		log.Error(err, "failed to update Component after ImageRepository creation", "ComponentName", component.Name)
+		return ctrl.Result{}, fmt.Errorf("error updating the component: %w", err)
+	}
+	log.Info("Component updated successfully, 'generate' annotation removed", "ComponentName", component.Name)
```
mmorhun, line 216: This code is duplicated. Should we 
```
if imageRepositoryFound == "" {
   <create image repository>
}
remove the annotation and update the component
```

controllers/component_image_controller.go
rcerven: I didn't remove it, but moved it before creating imagerepository 
https://github.com/konflux-ci/image-controller/pull/126/files#diff-6d10841540c4fbc83a67ba9f310705ea685b3650d4e865fbd8605c052209812aR178

controllers/component_image_controller.go
rcerven: bellow when component is updated, but I can add it to this log as well

controllers/component_image_controller.go
rcerven: yeah makes sense, will do

mmorhun (approved):

PR #148, opened 2024-09-29
Exclude Konflux application name from image name

tkdchen:
STONEBLD-2693


mmorhun (approved):

tnevrlka:
The reason behind this change wasn't clear to me until I've read the story.
 `Since Components have unique name, it's safe to remove application name from the image name`
However, upstream users don't have access to the `STONEBLD` Jira project. 
Could we start including more information in the commit message?

tkdchen:
@tnevrlka Good point. To save time from waiting the e2e-tests again, I updated the PR description this time. I also added an item for an improvement in the retro doc.

PR #149, opened 2024-09-29
Expose container port for probes

tkdchen:
kube-linter reports errors for this missing exposed container port.


tkdchen:
The issue is caught in PR check https://github.com/konflux-ci/image-controller/actions/runs/11069375671/job/30756800248?pr=147 and also happens from the `main` branch.

I'm using go1.22.7 and kube-linter@latest is used for the check in both GH action and local on my laptop.

mmorhun (approved):

PR #156, opened 2024-11-20
Prevent sequential underscores in quay robot account name

mmorhun:
When namespace has sequential `-` in its name, it converted to sequential `_` in quay robot account name which is not allowed and image repository provision fails. This PR prevents such error by removing duplicate `_` from robot account name.

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -850,7 +850,7 @@ func (r *ImageRepositoryReconciler) VerifyAndFixSecretsLinking(ctx context.Conte
 func generateQuayRobotAccountName(imageRepositoryName string, isPullOnly bool) string {
 	// Robot account name must match ^[a-z][a-z0-9_]{1,254}$
 
-	imageNamePrefix := imageRepositoryName
+	imageNamePrefix := removeDuplicateUnderscores(imageRepositoryName)
 	if len(imageNamePrefix) > 220 {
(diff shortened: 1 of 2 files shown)


MartinBasti:
how is conflict handled here?

what if there are 2 namespaces, `a_b` and `a__b` existing together ?

controllers/imagerepository_controller.go
```diff
@@ -867,6 +867,26 @@ func generateQuayRobotAccountName(imageRepositoryName string, isPullOnly bool) s
 	return robotAccountName
 }
 
+// removeDuplicateUnderscores replaces sequence of underscores with only one.
+// Example: ab__cd___e => ab_cd_e
+func removeDuplicateUnderscores(s string) string {
+	var result strings.Builder
```
MartinBasti, line 873: https://pkg.go.dev/regexp#Regexp.ReplaceAll can we use this instead?

mmorhun: Yes, we can. Replaced for loop implementation with regexp.

MartinBasti (approved):

controllers/imagerepository_controller.go
MartinBasti: thanks

MartinBasti:
Shouldn't be original commit just amended ?

rcerven (approved):

mmorhun:
@MartinBasti I'll do squash merge, no worries

PR #157, opened 2024-11-22
When removing ImageRepository, check if there is another ImageRepository

rcerven:
for the same repository, if there is, don't remove repository but only robot accounts for ImageRepository which is to be removed

[STONEBLD-2876](https://issues.redhat.com//browse/STONEBLD-2876)

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -118,8 +118,13 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 
 		if controllerutil.ContainsFinalizer(imageRepository, ImageRepositoryFinalizer) {
+			// Check if there isn't other ImageRepository for the same repository from other component
+			imageRepositoryFound, err := r.ImageRepositoryForSameUrlExists(ctx, imageRepository)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
 			// Do not block deletion on failures
-			r.CleanupImageRepository(ctx, imageRepository)
+			r.CleanupImageRepository(ctx, imageRepository, !imageRepositoryFound)
 
 			controllerutil.RemoveFinalizer(imageRepository, ImageRepositoryFinalizer)
 			if err := r.Client.Update(ctx, imageRepository); err != nil {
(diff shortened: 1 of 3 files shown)


controllers/imagerepository_controller.go
```diff
@@ -118,8 +118,27 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 
 		if controllerutil.ContainsFinalizer(imageRepository, ImageRepositoryFinalizer) {
+			// Check if there isn't other ImageRepository for the same repository from other component
```
mmorhun, line 121: I think it could be a separate func

controllers/imagerepository_controller.go
```diff
@@ -918,3 +928,26 @@ func generateDockerconfigSecretData(quayImageURL string, robotAccount *quay.Robo
...
+
+	imageRepositoryUrl := imageRepository.Status.Image.URL
+	imageRepositoryName := imageRepository.ObjectMeta.Name
+	imageRepositoryFound := false
+	for _, imageRepo := range imageRepositoriesList.Items {
+		if imageRepositoryUrl == imageRepo.Status.Image.URL {
+			if imageRepositoryName == imageRepo.ObjectMeta.Name {
+				continue
+			}
+			imageRepositoryFound = true
```
mmorhun, line 948: We don't need `imageRepositoryFound` variable at all, just return here

controllers/imagerepository_controller.go
```diff
@@ -918,3 +928,26 @@ func generateDockerconfigSecretData(quayImageURL string, robotAccount *quay.Robo
...
+		return false, err
+	}
+
+	imageRepositoryUrl := imageRepository.Status.Image.URL
+	imageRepositoryName := imageRepository.ObjectMeta.Name
+	imageRepositoryFound := false
+	for _, imageRepo := range imageRepositoriesList.Items {
+		if imageRepositoryUrl == imageRepo.Status.Image.URL {
+			if imageRepositoryName == imageRepo.ObjectMeta.Name {
+				continue
```
mmorhun, line 946: Would be good to have a comment here, like `Skipping original object`

controllers/imagerepository_controller.go
rcerven: done

controllers/imagerepository_controller.go
rcerven: done

mmorhun (approved):

mkosiarc (approved):

PR #163, opened 2025-01-08
Support skip the deletion of repositories

gbenhaim:
In order to support migration of users between clusters while preserving their repositories in Quay, support skipping the deletion of the repository in Quay by setting an annotation on the image repository resource.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,13 @@ After any successful operation, `status.message` is cleared.
 
 ---
 
+### Skip repository deletion
+
+By default, if the ImageRepository resource is deleted, the repository it created
+in Quay will get deleted as well.
+
+In order to skip the deletion of the repository in Quay, the `image-controller.appstudio.redhat.com/skip-repository-deletion` annotation should be set to "true".
+
 ## AppStudio Component image repository
 
 ### Image repository for Component builds
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -48,10 +48,11 @@ const (
 
 	ImageRepositoryFinalizer = "appstudio.openshift.io/image-repository"
 
-	buildPipelineServiceAccountName = "appstudio-pipeline"
-	updateComponentAnnotationName   = "image-controller.appstudio.redhat.com/update-component-image"
-	additionalUsersConfigMapName    = "image-controller-additional-users"
-	additionalUsersConfigMapKey     = "quay.io"
+	buildPipelineServiceAccountName      = "appstudio-pipeline"
+	updateComponentAnnotationName        = "image-controller.appstudio.redhat.com/update-component-image"
+	additionalUsersConfigMapName         = "image-controller-additional-users"
+	additionalUsersConfigMapKey          = "quay.io"
+	skipRepositoryDeletionAnnotationName = "image-controller.appstudio.redhat.com/skip-repository-deletion"
 )
 
 // ImageRepositoryReconciler reconciles a ImageRepository object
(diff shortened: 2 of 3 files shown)


controllers/imagerepository_controller.go
```diff
@@ -48,10 +48,11 @@ const (
...
 
-	buildPipelineServiceAccountName = "appstudio-pipeline"
-	updateComponentAnnotationName   = "image-controller.appstudio.redhat.com/update-component-image"
-	additionalUsersConfigMapName    = "image-controller-additional-users"
-	additionalUsersConfigMapKey     = "quay.io"
+	buildPipelineServiceAccountName      = "appstudio-pipeline"
+	updateComponentAnnotationName        = "image-controller.appstudio.redhat.com/update-component-image"
+	additionalUsersConfigMapName         = "image-controller-additional-users"
+	additionalUsersConfigMapKey          = "quay.io"
+	SkipRepositoryDeletionAnnotationName = "image-controller.appstudio.redhat.com/skip-repository-deletion"
```
chmeliik, line 55:
```suggestion
	skipRepositoryDeletionAnnotationName = "image-controller.appstudio.redhat.com/skip-repository-deletion"
```
For consistency with the other constants. I think it doesn't need to be exported.

controllers/imagerepository_controller.go
```diff
@@ -123,8 +124,17 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 			if err != nil {
 				return ctrl.Result{}, err
 			}
+
+			if imageRepositoryFound {
+				log.Info("Found another image repository for", "RepoURL", imageRepository.Status.Image.URL)
+			}
+
+			_, skipDeletion := imageRepository.Annotations[SkipRepositoryDeletionAnnotationName]
+			if skipDeletion {
```
chmeliik, line 133: We generally require the boolean-like annotations to be set to `"true"`, not just be present. Let's do the same here for consistency?

README.md
```diff
@@ -148,6 +148,13 @@ After any successful operation, `status.message` is cleared.
 
 ---
 
+### Skip repository deletion
+
+By default, if the ImageRepository resource is deleted, the repository it created
+in Quay will get deleted as well.
+
+In order to skip the deletion of the repository in Quay, the `image-controller.appstudio.redhat.com/skip-repository-deletion` should be added to the image repository resource.
```
chmeliik, line 156:
```suggestion
In order to skip the deletion of the repository in Quay, the `image-controller.appstudio.redhat.com/skip-repository-deletion` annotation should be set to `"true"`. 
```

controllers/imagerepository_controller_test.go
```diff
@@ -1226,6 +1226,73 @@ var _ = Describe("Image repository controller", func() {
...
+				Expect(organization).To(Equal(quay.TestQuayOrg))
+				Expect(imageRepository).To(Equal(customImageName))
+				return true, nil
+			}
+
+			deleteImageRepository(resourceKey)
+			// should not delete repository, because of the annotation
+			waitImageRepositoryGone(resourceKey)
+			Eventually(func() bool { return isDeleteRobotAccountInvoked }, timeout, interval).Should(BeTrue())
+			Eventually(func() bool { return isDeleteRepositoryInvoked }, timeout, interval).Should(BeFalse())
```
chmeliik, line 1288: I'm not sure if this really tests anything. `isDeleteRepositoryInvoked` starts out as `false`, so I think this `Eventually()` will just succeed immediately, even though the repository could still get deleted later

README.md
gbenhaim: done

controllers/imagerepository_controller.go
gbenhaim: done

controllers/imagerepository_controller.go
gbenhaim: done

controllers/imagerepository_controller_test.go
gbenhaim: the call to `deleteImageRepository(resourceKey)` will ensure that the image repository resources was deleted. It means that the call to remove the finalizer was made, and this happens after the function for deleting the external repository is called.

I removed the use of `Eventually` and cleaned the mock function for `DeleteRepositoryFunc`.   I hope it's less confusing now.

chmeliik: Ack, makes sense

chmeliik (approved):

gbenhaim:
thanks @chmeliik 
who can merge it?

chmeliik:
Yup, thanks for notifying :+1:

PR #169, opened 2025-02-12
Upgrade operator layout from v3 to v4

mmorhun:
This PR upgrades operator layout from v3 to v4.
New layout was generated and existing files were moved/merged to it.

Note, in layout v4, there is no `kube-rbac-proxy` image used.

More info on v3 vs v4 layout can be found in the kubebulder [docs](https://book.kubebuilder.io/migration/v3vsv4).

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -48,6 +48,10 @@ spec:
           - --health-probe-bind-address=:8081
         image: controller:latest
         name: manager
+        ports:
+        - containerPort: 8081
+          name: probes
+          protocol: TCP
         securityContext:
           allowPrivilegeEscalation: false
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Hard to tell, but it doesn't look like anything will break

If CI passes then LGTM

tnevrlka (approved):
Same as chmeliik, LGTM once CI passes

mmorhun:
If you are talking about Code Coverage check, then it's not really related to my changes. The check analyzes changed places, so if the coverage wasn't enough and you add an empty line - it will fail. In this PR, I haven't added any new functionality, it's just a migration of the operator scaffold.

PR #183, opened 2025-03-05
[ISV-5681] Cronjob to reset repository notifications

mantomas:
This cronjob is intended to regularly (once in a day initially) iterate over all Quay repositories  in a given namespace and reset all notifications failure counter to 0 if needed.
The job will reset counter for all notifications regardless if it is a push webhook defined within `ImageRepository` (the original reason for this cronjob to exist) or manually defined by repo user/owner. Making the job aware of the IR definitions was considered as a significant overhead for it's purpose.

Code tested successfully against owned namespace, all notifications were reset.

### DIFF
--- a/config/repository_notification_resetter/cronjob.yaml
+++ b/config/repository_notification_resetter/cronjob.yaml
@@ -10,7 +10,7 @@ spec:
       template:
         spec:
           containers:
-          - name: notification_resetter
+          - name: notification-resetter
             image: registry.redhat.io/rhel9/python-312:9.5-1739797362
             env:
               - name: QUAY_TOKEN
@@ -27,10 +27,10 @@ spec:
             command:
               - /bin/bash
               - '-c'
-              - python /notification_resetter/reset_notifications.py --namespace $(QUAY_NAMESPACE)
+              - python /notification-resetter/reset_notifications.py --namespace $(QUAY_NAMESPACE)
             volumeMounts:
               - name: notification-resetter-volume
-                mountPath: /notification_resetter
+                mountPath: /notification-resetter
             resources:
               limits:
                 cpu: 500m


.github/workflows/pr.yaml
```diff
@@ -68,3 +68,16 @@ jobs:
     - run: |
         cd config/registry_image_pruner/image_pruner/
         python3.9 test_prune_images.py
+
+  test_reset_notifications:
+    name: Run reset_notifications tests
```
mmorhun, line 73: I think we need to run it only if reset notifications script or job definition changed.

config/repository_notification_resetter/cronjob.yaml
```diff
@@ -0,0 +1,52 @@
...
+          containers:
+          - name: notification_resetter
+            image: registry.redhat.io/rhel9/python-312:9.5-1739797362
+            env:
+              - name: QUAY_TOKEN
+                valueFrom:
+                  secretKeyRef:
+                    name: quaytoken
+                    key: quaytoken
+              - name: NAMESPACE
```
mmorhun, line 21: nitpick: `NAMESPACE` might be confused with k8s namespace.

config/repository_notification_resetter/notification_resetter/reset_notifications.py
```diff
@@ -0,0 +1,181 @@
...
+        request = Request(
+            api_url,
+            headers={
+                "Authorization": f"Bearer {access_token}",
+            },
+        )
+
+        with urlopen(request) as resp:
+            if resp.status != 200:
+                raise RuntimeError(resp.reason)
```
mmorhun, line 148: Quay is sometimes unstable and can return weird codes or errors. We already had issues with such a behavior. Does it make sense to just log an error and continue processing other repos? (or maybe retry a few times in case of listing repos). But up to you.

mmorhun (commented):
Is it possible to add some metadata to notifications that we need to reset and do not reset everything? Would be it easy to implement?

.github/workflows/pr.yaml
mantomas: I copied this logic from previous task `test_prune_images` as I considered this as a best practice in this repo (run test for all PRs), but I can add some filtering to reduce resources.

mantomas:
> Is it possible to add some metadata to notifications that we need to reset and do not reset everything? Would be it easy to implement?

I think this would require a whole new logic to make the job 'aware' of all the `ImageRepository` definitions and I never worked with OpenShift API to estimate how complex this can be.
For our purpose, we can filter out notifications by title `SBOM-event-to-Bombino`, which will exclude all other notifications. This can be defined in cronjob template as env variable with list of titles of interest for future expansion.

I don't have any data for the notification use, but optionally we can leave it as is and add logging of the notification title for now, evaluate the job log after some time and decide based on that, if the additional logic is worth implementation.

.github/workflows/pr.yaml
mantomas: Added job to check code changes, so the tests will run only on code update.

.github/workflows/pr.yaml
```diff
@@ -68,3 +68,34 @@ jobs:
     - run: |
         cd config/registry_image_pruner/image_pruner/
         python3.9 test_prune_images.py
+
+  changes:
```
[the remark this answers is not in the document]
mantomas: Updated

.github/workflows/pr.yaml
```diff
@@ -68,3 +68,34 @@ jobs:
...
+      pull-requests: read
+    outputs:
+      notification_resetter: ${{ steps.filter.outputs.notification_resetter }}
+    steps:
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+        filters: |
+          notification_resetter:
```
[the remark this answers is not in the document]
mantomas: updated

.github/workflows/pr.yaml
```diff
@@ -68,3 +68,34 @@ jobs:
...
+  changes:
+    name: Check if notification resetter code changed
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      notification_resetter: ${{ steps.filter.outputs.notification_resetter }}
+    steps:
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v3
```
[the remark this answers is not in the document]
mantomas: I see your point - my thinking was a bit opposite, as the new `detect_changes_for_tests` job can be easily moved up and updated to be reused also with `test_prune_images`  (or any future logic), but this potential refactoring should be in separate PR from my POV.

config/repository_notification_resetter/notification_resetter/reset_notifications.py
```diff
@@ -0,0 +1,205 @@
...
+        # filter out when this is the case
+        rsp_message = json.loads(ex.read()).get("detail", "")
+        if ex.status == 400 and rsp_message.startswith(
+            "No repository notification found"
+        ):
+            LOGGER.info(
+                "Notification %s from %s/%s was not found", uuid, namespace, name
+            )
+        else:
+            raise (ex)
```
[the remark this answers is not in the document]
mantomas: Good point - replaced with proper logging so the job can proceed at least partially.

mmorhun (approved):

flacatus:
/retest

Recently toolchain operators updated the go version and broke the test. We update as well the go version and now should be fine

PR #191, opened 2025-04-08
Fix test reference to config/crd/bases

tnevrlka:
The controller directory got moved from `controllers` directory to `internal/controller` in 60686d3.

Test setup is expecting the `config/crd/bases` to be one directory above the test file, but due to the addition of the `internal` directory, it is now two directories above.

### DIFF
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -72,7 +72,7 @@ var _ = BeforeSuite(func() {
 
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{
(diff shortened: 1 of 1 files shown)


chmeliik (approved):

tnevrlka:
@rcerven Will the PR be ready for review anytime soon? If not, I would think about resolving the issue in this PR so that other PRs do not experience the check failure.

chmeliik:
+1 to fix it here. I don't think anyone could have noticed that #184 is fixing the problem, since the fix is hidden in the same commit that changes ~1000 lines

PR #192, opened 2025-04-09
fix(KFLUXSPRT-2659): always create a pull secret

arewm:
We currently only produce a pull secret if a component is linked. Changing the behavior to always create a pull secret so that a pull-only token can be shared with external parties if needed.

resolves: [KFLUXSPRT-2659](https://issues.redhat.com//browse/KFLUXSPRT-2659)


arewm:
I tried to race to push a fix for this issue first but @rcerven will be creating a new PR with a better fix.

PR #210, opened 2025-06-24
Automerge GitHub Actions updates

tnevrlka:
There is a lot of pull requests created for the repository that update dependencies.
Reviewing all of them manually takes time and usually isn't much more thorough than checking if the CI passes anyway.

Introduce a new package rule that automerges GH Actions.
GH Actions should be safe to update because the particular action would fail in the PR

Note: `allowedVersions` is changed from an array based on the recommendation from Renovate config lint

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,9 +8,14 @@
       "matchPackageNames": [
         "registry.access.redhat.com/ubi9/go-toolset"
       ],
-      "allowedVersions": [
-        "<2"
-      ]
+      "allowedVersions": "<2"
+    },
+    {
+      "description": "Automerge GitHub Actions updates",
+      "matchManagers": [
+        "github-actions"
+      ],
+      "automerge": true
     }
   ]
 }


tnevrlka:
After a discussion in a similar pull request in https://github.com/konflux-ci/build-service/pull/449, it was decided that Go dependencies should not be automerged.
I have updated the PR to only automerge GitHub Actions updates.

eskultety (approved):

renovate.json
```diff
@@ -8,9 +8,14 @@
       "matchPackageNames": [
         "registry.access.redhat.com/ubi9/go-toolset"
       ],
-      "allowedVersions": [
-        "<2"
-      ]
+      "allowedVersions": "<2"
+    },
+    {
+      "description": "Automerge GitHub Actions and Go dependencies updates",
```
tisutisu, line 14: nit pick: do we need to update the description after the review?

tnevrlka: Yes, thanks!

PR #215, opened 2025-08-08
Create application pull secret and link it to integration SA

rcerven:
Previously, we had a SA for each application. Now we will have only pull secret for application that will be linked to SA
"konflux-integration-runner" that will be in each namespace.

The new pull secret for application is an aggregated secret, that contains records for each image repository.

[STONEBLD-3611](https://issues.redhat.com//browse/STONEBLD-3611)

### DIFF
--- a/internal/controller/imagerepository_controller.go
+++ b/internal/controller/imagerepository_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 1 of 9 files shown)


internal/controller/imagerepository_controller.go
```diff
@@ -877,3 +898,233 @@ func (r *ImageRepositoryReconciler) ImageRepositoryForSameUrlExists(ctx context.
...
+
+	applicationPullSecretName := getApplicationPullSecretName(applicationName)
+	applicationPullSecret := &corev1.Secret{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: applicationPullSecretName, Namespace: namespace}, applicationPullSecret); err != nil {
+		log.Error(err, "failed to get application pull secret", "secretName", applicationPullSecretName)
+		return err
+	}
+
+	pullSecret := &corev1.Secret{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: pullSecretName, Namespace: namespace}, pullSecret); err != nil {
```
mmorhun, line 921: Do we really need to get the secret every time?

internal/controller/imagerepository_controller.go
```diff
@@ -877,3 +898,233 @@ func (r *ImageRepositoryReconciler) ImageRepositoryForSameUrlExists(ctx context.
...
+	if existingAuths.Auths == nil {
+		existingAuths.Auths = map[string]dockerConfigAuth{}
+	}
+
+	// If there are multiple pullsecrets for the same registry,
+	// keep the first one that was already present. Do not override unless explicitly requested (for rotation)
+	for reg, entry := range newAuths.Auths {
+		if _, authAlreadyExists := existingAuths.Auths[reg]; !authAlreadyExists {
+			existingAuths.Auths[reg] = entry
+			changed = true
```
mmorhun, line 962: Missing `brake`

internal/controller/imagerepository_controller.go
rcerven: well we need it's content to compare it with application secret

mmorhun (approved):

PR #221, opened 2025-08-27
don't revert image name in imageRepository upon change, but only add

rcerven:
message about it, to prevent argoCD to resync upon change

[STONEBLD-3746](https://issues.redhat.com//browse/STONEBLD-3746)

### DIFF
--- a/internal/controller/imagerepository_controller.go
+++ b/internal/controller/imagerepository_controller.go
@@ -60,7 +60,9 @@ const (
 	waitForRelatedComponentInitialWindowDuration  = 2 * 60
 	waitForRelatedComponentFallbackWindowDuration = 60 * 60
 
-	componentSaNamePrefix = "build-pipeline-"
+	componentSaNamePrefix                   = "build-pipeline-"
+	imageRepositoryNameChangedMessagePrefix = "Image repository name changed after creation"
+	imageRepositoryNameChangedMessageSuffix = "That doesn't change image name in the registry. To do that, delete ImageRepository object and re-create it with new image name."
 )
 
 // ImageRepositoryReconciler reconciles a ImageRepository object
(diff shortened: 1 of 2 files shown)


internal/controller/imagerepository_controller.go
```diff
@@ -277,17 +278,31 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-	// Do it here to avoid webhook creation.
+	// If it isn't set message
 	imageRepositoryName := strings.TrimPrefix(imageRepository.Status.Image.URL, fmt.Sprintf("quay.io/%s/", r.QuayOrganization))
 	if imageRepository.Spec.Image.Name != imageRepositoryName {
-		oldName := imageRepository.Spec.Image.Name
-		imageRepository.Spec.Image.Name = imageRepositoryName
-		if err := r.Client.Update(ctx, imageRepository); err != nil {
-			log.Error(err, "failed to revert image repository name", "OldName", oldName, "ExpectedName", imageRepositoryName, l.Action, l.ActionUpdate)
-			return ctrl.Result{}, err
+		imageNameDiffersMessage := fmt.Sprintf("%s from '%s' to '%s'", imageRepositoryNameChangedMessage, imageRepositoryName, imageRepository.Spec.Image.Name)
```
chmeliik, line 284: > Image repository name changed after creation from '%s' to '%s'

Which means the name changed in the spec, but the controller ignored that, right? Can we mention that in the message?

Maybe something like this:

> Image repository name changed after creation from '%s' to '%s'. This does not change the name of the the actual image repository in the registry. To do that, delete this ImageRepository object and re-create it with a different repository name.

mmorhun (approved):
Please improve the message as Adam has suggested.

internal/controller/imagerepository_controller.go
rcerven: improved message

PR #224, opened 2025-09-12
upon application removal, remove application pull secret from namespa…

rcerven:
…ce SA

[STONEBLD-3777](https://issues.redhat.com//browse/STONEBLD-3777)

### DIFF
--- a/internal/controller/application_controller.go
+++ b/internal/controller/application_controller.go
@@ -36,7 +36,10 @@ import (
 	appstudioredhatcomv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 )
 
-const NamespaceServiceAccountName = "konflux-integration-runner"
+const (
+	IntegrationTestsServiceAccountName = "konflux-integration-runner"
+	ApplicationSecretLinkToSaFinalizer = "application-secret-link-to-integration-tests-sa.appstudio.openshift.io/finalizer"
+)
 
 // dockerConfigJson represents the structure of a .dockerconfigjson secret
 type dockerConfigJson struct {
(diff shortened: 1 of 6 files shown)


internal/controller/application_controller.go
```diff
@@ -326,3 +352,53 @@ func (r *ApplicationPullSecretCreator) doesApplicationPullSecretExist(ctx contex
...
+	err := r.Client.Get(ctx, types.NamespacedName{Name: NamespaceServiceAccountName, Namespace: namespace}, serviceAccount)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		log.Error(err, "failed to read namespace service account", l.Action, l.ActionView)
+		return err
+	}
+
+	unlinkSecret := false
```
mmorhun, line 370: Better name it `shouldUpdate` or something like that.

internal/controller/application_controller.go
```diff
@@ -36,7 +36,10 @@ import (
 	appstudioredhatcomv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 )
 
-const NamespaceServiceAccountName = "konflux-integration-runner"
+const (
+	NamespaceServiceAccountName = "konflux-integration-runner"
```
mmorhun, line 40: Name is not clear. We can have sevaral namespace-wide Service Accounts. Please be more specific.

internal/controller/application_controller.go
```diff
@@ -36,7 +36,10 @@ import (
...
+	ApplicationFinalizer        = "appstudio.openshift.io/application"
```
mmorhun, line 41: Please use more specific name for the finalizer.

internal/controller/application_controller.go
```diff
@@ -326,3 +352,53 @@ func (r *ApplicationPullSecretCreator) doesApplicationPullSecretExist(ctx contex
 
 	return true, nil
 }
+
+// unlinkApplicationSecretFromServiceAccount ensures that the given secret is not linked with the namespace service account.
+func (r *ApplicationPullSecretCreator) unlinkApplicationSecretFromServiceAccount(ctx context.Context, secretNameToRemove, namespace string) error {
```
mmorhun, line 357: The name is quite confusing, since it says `*FromServiceAccount`, which implies generic case. However, you operate on `NamespaceServiceAccountName` SA only.

internal/controller/application_controller.go
rcerven: ok

internal/controller/application_controller.go
rcerven: ok

mmorhun (approved):

PR #294, opened 2026-04-20
Update notification-resetter memory resources

mantomas:
Number of handled repositories raised and the cronjob is randomly OOM killed. The memory was set very low and might fall into unsafe `limit` range with higher OOM kill risk - this is to remediate.

Relates to [KONFLUX-13244](https://redhat.atlassian.net/browse/KONFLUX-13244)

[KONFLUX-13244]: https://redhat.atlassian.net/browse/KONFLUX-13244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/repository_notification_resetter/cronjob.yaml
+++ b/config/repository_notification_resetter/cronjob.yaml
@@ -34,10 +34,10 @@ spec:
             resources:
               limits:
                 cpu: 500m
-                memory: 1024Mi
+                memory: 768Mi
               requests:
                 cpu: 150m
-                memory: 512Mi
+                memory: 768Mi
             securityContext:
               readOnlyRootFilesystem: true
           restartPolicy: OnFailure
--- a/config/repository_notification_resetter/cronjob.yaml
+++ b/config/repository_notification_resetter/cronjob.yaml
@@ -34,10 +34,10 @@ spec:
             resources:
               limits:
                 cpu: 500m
-                memory: 768Mi
+                memory: 1024Mi
               requests:
                 cpu: 150m
-                memory: 768Mi
+                memory: 1024Mi
             securityContext:
               readOnlyRootFilesystem: true
           restartPolicy: OnFailure
--- a/config/repository_notification_resetter/cronjob.yaml
+++ b/config/repository_notification_resetter/cronjob.yaml
@@ -34,10 +34,10 @@ spec:
             resources:
               limits:
                 cpu: 500m
-                memory: 512Mi
+                memory: 1024Mi
               requests:
                 cpu: 150m
-                memory: 128Mi
+                memory: 512Mi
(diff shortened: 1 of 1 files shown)


mantomas:
@mmorhun can you please take a look?

config/repository_notification_resetter/cronjob.yaml
```diff
@@ -34,10 +34,10 @@ spec:
             resources:
               limits:
                 cpu: 500m
-                memory: 512Mi
+                memory: 1024Mi
               requests:
                 cpu: 150m
-                memory: 128Mi
+                memory: 512Mi
```
rcerven: set requests and limits the same, it is safer

mantomas: Good point - I set it somewhere in between for both request/limit as the 512 might not be enough in some environments.

rcerven: just set it to 1024 for both so you don't have to bump it any time soon

mantomas: done

mantomas: @rcerven anything I can do to make this released? Can this config be merged without passing tests as it is failing everywhere?

mmorhun (changes requested):
If we do this change here, all Konflux deployments will be effected including dev, test and upstream.
We try to use minimal required amount fo resources, so Konflux can be deployed on local cluster.
If you target specific infra, please update its configuration, e.g. [here](https://github.com/redhat-appstudio/infra-deployments/tree/main/components/image-controller).

mantomas:
@mmorhun I'm not sure I understand the linked configuration. If this change is problematic I don't want to create issues by changing the default values here - can someone from image-controller maintainers take the [KONFLUX-13244](https://redhat.atlassian.net/browse/KONFLUX-13244) ticket and update the affected environments (6 clusters mentioned in the ticket)?

[KONFLUX-13244]: https://redhat.atlassian.net/browse/KONFLUX-13244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mmorhun:
@mantomas you need to patch stage and prod envs, see [example](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/image-controller/production/base/pruner_cronjob_resources_patch.yaml).

mantomas:
Closing in favor of change in infra-deployments.

PR #311, opened 2026-05-14
Configure golangci-lint and fix linter errors

mmorhun:
Configure golangci-lint and fix linter errors.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -285,7 +285,11 @@ func readConfig(envVarName string, path string) (string, error) {
 	}
 
 	if configValue == "" && path != "" {
-		fileInfo, err := os.Stat(path)
+		// There is no user-controlled input that could cause path traversal.
+		// The path is constructed filepath.Join(mountPoint, QuayApiFileName),
+		// where mountPoint is either a default constant or an environment variable.
+		// #nosec G703
+		fileInfo, err := os.Stat(path) // nolint:gosec
 		if err != nil {
 			if errors.Is(err, os.ErrNotExist) {
 				// No config provided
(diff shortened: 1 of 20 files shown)


cmd/main.go
```diff
@@ -405,7 +409,10 @@ func buildQuayHttpClient() (*http.Client, error) {
 		setupLog.Info("Additional CA certificate loaded successfully")
 	}
 
-	return &http.Client{Transport: transport}, nil
+	return &http.Client{
+		Transport: transport,
+		Timeout:   time.Minute,
```
MartinBasti, line 416: Why this change?

mmorhun: To be safe, the default timeout is unlimited.
Also, linter asked for it in quay client code, but messing with context per request is not what we want to do, at least now, so fixing it in one place.

pkg/quay/quay.go
```diff
@@ -72,29 +72,30 @@ func NewQuayClient(c *http.Client, authToken, url string) *QuayClient {
...
 
-func (r *QuayResponse) GetJson(obj interface{}) error {
-	defer r.response.Body.Close()
-	body, err := io.ReadAll(r.response.Body)
-	if err != nil {
-		return fmt.Errorf("failed to read response body: %s", err)
-	}
-	if err := json.Unmarshal(body, obj); err != nil {
-		return fmt.Errorf("failed to unmarshal response body: %s, got body: %s", err, string(body))
+func (c *QuayResponse) GetJson(obj any) error {
```
rcerven: can we keep it as 'r'  ?    (r *QuayResponse)

so it is not the same as QuayClient

mmorhun: ok

MartinBasti (approved):

rcerven (approved):

PR #314, opened 2026-05-20
fix: protect RepositoryTimesForMetrics from concurrent map access

MartinBasti:
The global map was read/written/deleted from multiple reconciler goroutines without synchronization, which can cause a fatal panic. Replace the exported map with mutex-protected helper functions.

### DIFF
--- a/internal/controller/imagerepository_controller.go
+++ b/internal/controller/imagerepository_controller.go
@@ -86,10 +86,7 @@ func (r *ImageRepositoryReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func setMetricsTime(idForMetrics string, reconcileStartTime time.Time) {
-	_, timeRecorded := metrics.RepositoryTimesForMetrics[idForMetrics]
-	if !timeRecorded {
-		metrics.RepositoryTimesForMetrics[idForMetrics] = reconcileStartTime
-	}
+	metrics.SetRepositoryTimeIfAbsent(idForMetrics, reconcileStartTime)
 }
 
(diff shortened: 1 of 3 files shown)


MartinBasti:
More context from AI
```
Yes, this is a legitimate fix. Here's the analysis:

  The bug is real: RepositoryTimesForMetrics is a plain map[string]time.Time accessed without any synchronization from the Reconcile method.
  In Go, concurrent reads and writes to a map cause a fatal runtime panic (not just a data race — the runtime detects it and crashes the
  process).
  
  The codebase does use concurrent goroutines that hit this map. Even though SetupWithManager doesn't explicitly set MaxConcurrentReconciles,
  the controller-runtime default is 1 worker per controller. However, there are 3 separate controllers registered in cmd/main.go:213-232, and
  the ImageRepositoryReconciler.Reconcile method is called from the controller-runtime's worker goroutine pool. More importantly:

  1. controller-runtime's default of 1 concurrent reconcile only guarantees that one reconcile runs at a time for that specific controller —
  but the map is a package-level global, so if any other code path (or a future bump to >1 worker) touches it, you get a crash.
  2. Even with a single worker, the go func() goroutine in pkg/metrics/metrics.go:64 runs concurrently and could theoretically interact with
  the same package's state.
  3. The Go race detector would flag this regardless of actual concurrency — it's a correctness issue by Go's memory model.

  The fix is clean and correct: wrapping the map in a sync.Mutex with Set/Get/Delete helpers is the standard approach. Using sync.Mutex (not
  sync.RWMutex) is fine here since operations are cheap and contention is low. The test includes a proper concurrency stress test
  (TestRepositoryTimeConcurrency).

```

PR #330, opened 2026-06-15
Replace expired CA cert in tests

mmorhun:
## Description

Replaces expired CA cert in `main_test.go` unit test.

## Testing

`make test`


cmd/main_test.go
```diff
@@ -390,22 +390,22 @@ func Test_buildQuayHttpClient(t *testing.T) {
 		g.RegisterTestingT(t)
 
 		const caCert = `-----BEGIN CERTIFICATE-----
```
chmeliik, line 392: Optionally, it would be nice to add a comment about how this CA cert was generated in case we need to do it again in the future

chmeliik (approved):

cmd/main_test.go
mmorhun: It's just a CA cert used it tests, nothing special about it. Any valid CA cert would work.

chmeliik: Ok, probably no need then

PR #326, opened 2026-06-04
feat: implement dual-group support for ImageRepository API migration

rcerven:
feat: implement dual-group support for ImageRepository API migration

Introduce new ImageRepository CRD under konflux-ci.dev/v1alpha1 API group alongside
existing appstudio.redhat.com/v1alpha1, enabling both API groups to coexist
during gradual migration.

Dual Controller Infrastructure:
- Created two separate reconciler instances (imagerepository-konflux and imagerepository-appstudio)
- Added IsOldGroup field to ImageRepositoryReconciler to distinguish controller instances
- Implemented bidirectional conversion functions (convertAppstudioToKonflux, convertKonfluxToAppstudio)
- Added group-aware update helpers (updateImageRepository, updateImageRepositoryStatus)
- Both controllers share same Reconcile() logic but operate on different API groups

Critical Migration Safety Fixes:
- Enhanced ImageRepositoryForSameUrlExists to check BOTH API groups during migration
- Prevents Quay repository deletion when ImageRepositories from different groups reference same repo
- Old group controller continues handling application pull secret operations for backward compatibility

Secret Naming Differentiation:
- New group (konflux-ci.dev): -img-push / -img-pull suffixes
- Old group (appstudio.redhat.com): -image-push / -image-pull suffixes
- Different suffixes prevent secret name collisions and separate ownership
  when both groups coexist on the same cluster during migration

New Group Field Immutability:
- Added CEL validation to spec.image.name in konflux-ci.dev/v1alpha1:
  MaxLength=2048, pattern validation, and XValidation rule preventing
  modification after initial creation ("Image repository name cannot be changed")
- ProvisionImageRepository now only populates spec.image.name when empty,
  preserving user-provided names; full qualified name appears only in status.image.url
- Controller-side name-change detection (status message warning) is now
  old-group-only (gated by r.IsOldGroup); new group relies on CEL enforcement
- Updated new-group tests: replaced "should add message when image name was edited"
  with rejection assertion using MatchError(ContainSubstring(...))

Test Infrastructure:
- Created imagerepository_controller_old_test.go with full test coverage for old API group
- Duplicated test helper functions with _old suffix (createImageRepository_old, getImageRepository_old, etc.)
- Both API groups registered in test scheme for dual-group testing scenarios
- Updated test certificate with 10-year validity (expires 2036) to prevent expiration issues

Build & Test Improvements:
- Updated manifests target to scan both api/ and internal/controller/ for RBAC markers
[...]

PR #338, opened 2026-06-28
Fix .codecov.yml: ignore generated files and config manifests

eedri:
## Summary

- Add `ignore:` patterns for generated files (`zz_generated*.go`, `config/**`, `vendor/**`) that were dragging down patch coverage metrics
- Keeps existing coverage targets (80% patch, auto project) and informational mode
- Keeps `comment: false` per team preference

## Context

The Konflux Code Coverage Report flagged image-controller at **42.25% patch coverage** over the last 7 days. Investigation showed that generated deepcopy files (`zz_generated.deepcopy.go`, +257 lines) and CRD manifests were being counted as uncovered production code, significantly skewing the metric.

## Test plan

- [ ] Verify Codecov stops counting `zz_generated*.go` and `config/` files in patch coverage

Signed-off-by: Eyal Edri <eedri@redhat.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,11 @@
 codecov:
   require_ci_to_pass: false
 
+ignore:
+  - "**/zz_generated*.go"  # generated deepcopy code
+  - "config/**"            # CRD, RBAC, and kustomize manifests
+  - "vendor/**"            # vendored dependencies
+
 coverage:
   status:
     project:
@@ -13,4 +18,6 @@ coverage:
         target: 80%
         informational: true
 
-comment: false
+comment:
+  layout: "reach,diff,flags,files,footer"
+  require_changes: true
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -18,6 +18,4 @@ coverage:
         target: 80%
         informational: true
 
-comment:
-  layout: "reach,diff,flags,files,footer"
-  require_changes: true
+comment: false


chmeliik (changes requested):
> Enable PR comments so developers see coverage impact before merging (was comment: false)

We explicitly asked not to do this in https://github.com/konflux-ci/image-controller/pull/328#issuecomment-4731595111

eedri:
Hi, I sent this following https://github.com/konflux-ci/image-controller/pull/326, where low coverage was detected on the feature PR, can you verify that without this comment the team is aware of the patch coverage going in on each PR? 

Right now the repo patch coverage is on 42.25% for last week, if we say that we don't need that comment in order to maintain the 80% target, i'm OK to remove it.

chmeliik:
> Hi, I sent this following #326, where low coverage was detected on the feature PR, can you verify that without this comment the team is aware of the patch coverage going in on each PR?

Speaking for myself, I don't see value in chasing coverage numbers in each PR. The number is one thing, actual valuable coverage another.

That said, if the comment is the only way to see the number, I guess we can live with a bit more spam. Between Qodo and Fullsend, Codecov will be one of the less noisy spammers anyway. But we will need more team members' opinions on this.

eedri:
I honestly don't mind if we add comemnt or not, what we're aiming it to have full patch coverage above 80%, otherwise we'll always be chasing code coverage on the repo and add it post mortem.

I also don't understand how the feature was truely tested or gated if it wasn't merged with tests, are the tests exists in other repos?

rcerven:
https://github.com/konflux-ci/image-controller/pull/326 is part of effort of new model and new group API, which still needs 3 PRs to be complete

we have so far only unit tests and not e2e tests for the feature (as it isn't fully done)

as for comments with % of coverage, I don't think it adds value, we are always adding new unit tests to cover new things

ofc in this case coverage was lower because CRD and generated files were added as well

eedri:
can't we aim to merge a piece of feature with full coverage? so you can verify if you have any errors or regression on the PR level rather than adding the e2e tests at a later stage?, we all want to get to point where we can release any time with any PR, and having code that is fully tested decouples that connection between multiple PRs on the same feature.

Anyway, I'll drop the comment part if that is bothering the team, but it doesn't mean we're going to stop aiming for 80% patch coverage, at some point we might want to fail CI on it as well ( some teams already do it).

chmeliik (approved):

